PR #14, opened 2015-10-27
Thrift: initial advice

rrees:
Commas are optional in the Thrift definition and we have started omitting them in our definitions.

The optional versus required is personal opinion but looking at how difficult it is to backfill database fields I thought it would be good to get this in early.

### DIFF
--- a/thrift.md
+++ b/thrift.md
@@ -1,0 +1,5 @@
+# Thrift
+
+Don't use commas in the structure definitions.
+
+Don't mark a field as *required* unless the format is not yet in production.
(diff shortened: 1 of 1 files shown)


LATaylor-guardian:
Semi colons are also optional and sometimes used as commas are at the end of a field declaration.

LATaylor-guardian:
Ordering of fields is is also important and should not be changed once available.

rrees:
Would anyone really do that?

LATaylor-guardian:
Always worth being explicit. Not everyone has Thrift experience.

philwills:
:+1:

PR #15, opened 2016-01-05
Security guidelines

adamnfish:
Some recommendations / guidelines relating to securing our applications. These are all based on real issues we've faced and will probably be an evolving list as we find new areas of concern and best practices move on.

### DIFF
--- a/security.md
+++ b/security.md
@@ -1,0 +1,54 @@
+# Security
+
+## Sensitive data
+
+**Before building an application think about what (if any) sensitive
+data is involved.** Be particularly aware of any personal user data
+(including email addresses).
+
+**Have a plan for storing sensitive data safely.** Make sure the ways
+that data can be accessed are well understood. How will you ensure the
+data remains safe?
+
+**Limit access to sensitive data.** Only people and services that
+directly depend on the data should be able to access it.
+
+**Keep sensitive data to yourself.** Keep your team's sensitive data
+within the team and under your control. Don't share it with other
+teams, don't store it in cookies and don't let it sit in caches.
+
+## Application development
+
+**Keep secrets out of repositories.** Application secrets must not be
+stored in version control, even if the repo is "private". To prevent
+accidental leakage, files containing secrets should be stored away
+from the project, completely outside of the source tree.
+
+**Use the OWASP Security Knowledge Framework**. Before releasing a new
+application make sure you run through the Security Knowledge Framework
+to get pointers on areas that might need extra attention. Revisit it
+from time to time to stay on top of changes in the application and the
+team.
+
+## AWS
+
+**Do not use permanent AWS credentials.** Applications running in AWS
+should use `assumeRole` to gain access to other AWS resources. For
+local development you should use AWS profiles to manage credentials
+and these should be temporary credentials (e.g. provided by Janus).
+
+## HTTPS everywhere
+
+**_Everything_ should be served over HTTPS.**
+
+## Personal hygiene
+
+**Disk encryption.** All machines that you do work on must have disk
+encryption enabled.
+
+**2FA.** Whenever a service offers 2FA, you should enable it.
+
+**Use a password manager**. Don't reuse simple passwords across
+services. A password manager is a good way to help use strong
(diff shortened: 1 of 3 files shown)


rrees:
:+1: HTTPS everywhere is going to require a lot infrastructure changes though

security.md
```diff
@@ -0,0 +1,54 @@
...
+to get pointers on areas that might need extra attention. Revisit it
+from time to time to stay on top of changes in the application and the
+team.
+
+## AWS
+
+**Do not use permanent AWS credentials.** Applications running in AWS
+should use `assumeRole` to gain access to other AWS resources. For
+local development you should use AWS profiles to manage credentials
+and these should be temporary credentials (e.g. provided by Janus).
```
theefer: I might be wrong, but I suspect this is one of the recommendations which currently most projects do **not** follow. Are there many practical examples of this, in particular the Janus-only DEV credentials?

adamnfish:
@rrees not sure I agree. We already do this naturally for pretty much every API and website we build, I'd be interested to hear IRL if you can think of any exceptions to this? It does need clarifying that I don't necessarily include traffic inside an AWS account (i.e. load balancer -> instances) here.

theefer:
I do think there's room to automate SSL cert management a lot more (Let's Encrypt?) than the current manual process with cert-magic and emails to Enterprise, in order for it to be much less of a burden when dealing with lots of smaller services/projects.

security.md
adamnfish: Yep, there are quite a few examples of where this isn't done at the moment. We're slowly moving existing projects away from permanent credentials (particularly in the tools team which has multiple accounts) so there are quite a few examples in there. I think Lindsey and Simon would be best placed to point you at their examples? More than the existing stuff though, this recommendation is important for teams that are thinking about how to design new applications.

adamnfish:
@theefer yep, hopefully this will encourage us to solve those problems in a better way. It's certainly already possible because pretty much everything we do is already secure, but there's definitely some tasty fruit that's tantalisingly in reach here.

security.md
```diff
@@ -0,0 +1,54 @@
...
+
+**Disk encryption.** All machines that you do work on must have disk
+encryption enabled.
+
+**2FA.** Whenever a service offers 2FA, you should enable it.
+
+**Use a password manager**. Don't reuse simple passwords across
+services. A password manager is a good way to help use strong
+passwords. The `pwgen` command can be used to generate good passwords,
+especially with the `-s` and `-y` flags (`pwgen -sy <length>`).
```
satterly: To install use `brew install pwgen` on a Mac.

security.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Application development
+
+**Keep secrets out of repositories.** Application secrets must not be
+stored in version control, even if the repo is "private". To prevent
+accidental leakage, files containing secrets should be stored away
+from the project, completely outside of the source tree.
+
+**Use the OWASP Security Knowledge Framework**. Before releasing a new
+application make sure you run through the Security Knowledge Framework
```
satterly: Add a link? I assume you mean these guys ... https://www.securityknowledgeframework.org/

adamnfish: I wrote this a couple of days ago, before this link existed https://skf.gutools.co.uk/ Now that it works you're absolutely right. There'll be more info to follow on this subject.

security.md
adamnfish: I'm not sure "how to install software on your dev machine" belongs in the security section of the recommendations repo. I reckon people will work this out but maybe if a few mac users feel like this is necessary I can add a note to that effect?

rrees:
@adamnfish I thought this did refer to HTTPS within accounts. I agree that it is a sensible rule for external termination.

cb372:
Maybe add an AWS-specific note reminding people that they should make rules (e.g. IAM policies, security group rules) as restrictive as possible, only widening them later when the need arises. It's often tempting to do the opposite, leaving everything open for the sake of developer convenience, then not getting around to tightening them up later.

philwills:
I'm going to merge this, as I'd rather have it in that not and we can improve on it later.

PR #18, opened 2016-03-03
Updates to security document

philwills:
Based on comments on previous PR.

cc @adamnfish


adamnfish:
:+1:  Nice one @philwills. I keep forgetting to re-do this, thanks for the fixes and the prompt.

PR #19, opened 2016-03-22
Update scala and add dependency injection warning

mchv:


scala.md
```diff
@@ -5,6 +5,7 @@ Scala
...
+ * If some dependencies are unavailable for the latest version, it is acceptable to use the latest patch version of the previous minor version but try to avoid it. At the time of writing 2.11.8 is the latest version, so 2.10.4 may be acceptable. If you have a dependency that isn't available for one of these versions, it's effectively dead and you should plan how to move away from it. 
  * If you come to a codebase using an older version of Scala, it should be upgraded before any new features are added.
- * Run on a [version of the JVM with public updates](http://www.oracle.com/technetwork/java/eol-135779.html#javase). At the time of writing this means JDK7 or 8.
+ * Run on a [version of the JVM with public updates](http://www.oracle.com/technetwork/java/eol-135779.html#javase). At the time of writing this means JDK8 or 9.
+ * Avoid dependency injection in general. Most of the time, dependency injection does not solve a real problem. You don't need to have a single place where you define which concrete instance you trait should be used. If you have to use dependency injection always use conpile time dependency injection.
```
rrees: "your trait"

"compile time"

Link to what you mean by "compile-time dependency injection".

philwills:
I think you know, it's a :+1: from me

davidfurey:
Avoiding dependency injection seems like a fair recommendation to me, as does recommending compile-time over run-time.  

Given that it is the Play framework that has prompted this discussion, I think we should also have a recommendation that is specific to the Play framework.

tomverran:
I'd suggest altering the wording to specifically explain that you're suggesting the avoidance of dependency injection _frameworks_ as opposed to dependency injection as a general principle.

Or am I wrong in thinking that? I must confess that I haven't really got my head around how to structure applications without sticking together classes, either manually or otherwise.

johnduffell:
Looks great, although probably the biggest problem with these recommendations from my point of view is that I've never read them :blush:

mchv:
> I'd suggest altering the wording to specifically explain that you're suggesting the avoidance of dependency injection frameworks as opposed to dependency injection as a general principle.

sounds sensible, I will update the wording.

philwills:
Just noticed this hanging about. Unless I hear otherwise, I'm going to merge it shortly.

johnduffell:
good idea but @mchv never did update the wording as he planned

philwills:
Yea, I was going to try and do that myself, but then realised I'm an old curmudgeon who actually prefers them as @mchv wrote them.

mchv:
I have updated the wording to be explicit about why **dependency injection** should be avoided.

scala.md
```diff
@@ -5,6 +5,19 @@ Scala
...
+
+Instead of `injecting` your dependencies, you should simply [pass them as parameters](http://debasishg.blogspot.com/2011/03/pushing-envelope-on-oo-and-functional.html) of functions you are calling. This enable [referential transparency](https://en.wikipedia.org/wiki/Referential_transparency) of your programs.
+
+Most of the time, dependency injection does not solve a real business problem. You don't need to have a single place where you define which concrete instance your trait should be used. This is`accidental complexity`, as coined by Fred Brooks in [No Silver Bullet](`https://en.wikipedia.org/wiki/No_Silver_Bullet`)  
+
+With [The play framework](https://www.playframework.com/) you should always use [compile-time dependency injection](https://www.playframework.com/documentation/2.5.x/ScalaCompileTimeDependencyInjection) which refers to using the [cake pattern](http://www.cakesolutions.net/teamblogs/2011/12/15/dependency-injection-vs-cake-pattern), a object oriented way to specify your components declaratively in scala.
```
philwills: We're not actually advocating cake here are we?

johnduffell: that would certainly be very controversial, in frontend we are using macwire although I have no idea whether that's the right thing to do (it confuses me a lot)
However @alexduf spent a while looking into it so his input would probably be enlightening

mchv: just created #22 to not confuse people

philwills: I've only done a tiny project, but for that https://github.com/guardian/audience-platform/blob/master/audience-explorer/app/AppLoader.scala is about the level of complexity I can deal with.

I'm not convinced that the Routes generation is worth it when you have to integrate with it though, which makes me start to think heretical thoughts, like why not construct something using https://github.com/finagle/finch, https://github.com/travisbrown/circe, http://eed3si9n.com/gigahorse/ and Twirl and avoid the Play magic completely. 

Half magic seems like the worst of both worlds.

alexduf: @johnduffell I used macwire mainly because there was a lot of wiring to do, the task being big enough that I would actually gain time by using that macro. Now if it's getting people confused (which is clearly not the goal) we can start wiring manually and remove it?
@philwills The whole route generation is the main reason I looked into macwire. Agreed it's combining half magic with even more half magic, but it has the advantage of being able to wire a Router without worrying about parameter order.

But just to be clear, I'm not saying macwire is a silver bullet, I'm very conscious of its limits.

Issue #28, opened 2017-07-25
resiliency


davidfurey:
The resilience levels in resiliency.md are interesting, but is this structure something we use?  If we do, then great.  Otherwise I think we should either start using it or delete it.  Thoughts?

philwills:
No, I don't think it's something we ever fully adopted. I think it's good to define that there are different concerns which are applicable for different services, but I'd agree that this doesn't represent reality. I'd rather replace than delete, but maybe that's a sensible first step.

PR #29, opened 2017-07-25
Remove gold/silver/bronze resiliency

davidfurey:
We don't use those terms

(see https://github.com/guardian/recommendations/issues/28)


tackley:
That grid was a beautiful fantasy that seemed a great idea at the time :)

The moral, of course, is that just writing something down and doing nothing else (as I did in this case) doesn't change anything...

PR #31, opened 2017-11-17
Focus recommendations around injection

adamnfish:
Remove irrelevant detail to focus on the guidance. (We get it)


mchv:
I think the details are useful just because I have got asked the reasons why it is bad many times. I understand that for a more senior developer, this may sounds unnecessary and even considered noise polluting the main objective of the recommendations.

markjamesbutler:
@mchv @adamnfish did we reach a conclusion on this? Personally, I think the detail could be useful for devs. Although it could be moved to a separate location.

PR #34, opened 2019-04-25
Config recommendations

tomrf1:


config.md
```diff
@@ -0,0 +1,18 @@
...
+1. Granular read/write access
+2. Free to use
+3. Versioned
+4. Encryption at rest
+
+##### Scala config
+Our own [simple-configuration](https://github.com/guardian/simple-configuration) library provides a convenient way to load config from Parameter Store and into [Typesafe config](https://github.com/lightbend/config) objects. The readme includes an example of using it in a Play app.
+
+For existing projects which currently have their config stored in Typesafe (HOCON) format files, you can migrate to Parameter Store using the [parameter-store-migration](https://github.com/guardian/parameter-store-migration) tool.
+
```
tomrf1: I haven't decided if this section should say anything about application.conf / reference.conf.
Possible recommendations could be:
1. just put everything in PS only
2. list *all* config items in the application.conf for the sake of documentation, and set environment-specific items to `null`
3. put default values in reference.conf
4. it's up to developers to decide what they prefer...

config.md
```diff
@@ -0,0 +1,19 @@
+### Config
+
```
jfsoul: Could add a disclaimer here: generally we could teams to pick the right technology for the task but there are advantages to consistency for config... etc etc

config.md
rupertbates: I like having any non sensitive config in application.conf for a site or reference.conf for a library. That makes it really easy to see what values are when debugging. Also by putting values in reference.conf for libraries, any application which uses that library can just use the defaults without having to provide its own values.

stettix: I think putting defaults in `reference.conf` is good practice, and leaving environment specific and sensitive configs for `application.conf` or PS.

JonNorman: On a recent project I have written all of my config values in `application.conf` and set to `null` any sensitive keys that are set by PS. I think this works well for documenting the requirements for the application.

I haven't used `reference.conf`, but not for any reason other than I haven't had time to make the split. It sounds like a good idea.

PR #37, opened 2020-08-19
Remove outdated or harmful rules

gtrufitt:
## What does this change?

This is a first step in updating these recommendations, by removing outdated or harmful rules for modern browsers.

For example "Minimize the number of HTTP requests using methods such as bundling to avoid latency issues." is less relevant (and even harmful to cache-ability) in HTTP/2 supporting browsers.

I have not added or majorly edited any recommendations.

### DIFF
--- a/client-side.md
+++ b/client-side.md
@@ -77,3 +77,5 @@ various areas below.
 - Define web performance budgets and track them
 
 ### Build tools/task runners
+
+- Have an automated process to produce your final assets (e.g. webpack, rollup, microbundle ).
--- a/client-side.md
+++ b/client-side.md
@@ -28,7 +28,7 @@ free to justify why not.
 
 - If starting a new project, favour ES6 modules unless you can justify
   otherwise.
-- Do not commit your dependencies into Git. Instead, lock your dependencies (direct and transitive) using a lockfile.
+- Prefer locking your dependencies (direct and transitive) using a lockfile rather than committing them into Git.
 
 ## Strategies
 
--- a/client-side.md
+++ b/client-side.md
@@ -6,9 +6,6 @@ free to justify why not.
 
 ## Assets
 
-- Pre-process before serving in production (e.g. Sass, ES6).
-- Minimize the number of HTTP requests using methods such as bundling
-  to avoid latency issues.
 - Minify assets (e.g. uglify).
 - Gzip all textual assets served, using GZip level 6 where possible
 - Optimise images for size (e.g. jpegtran, pngquant, giflossy, svgo,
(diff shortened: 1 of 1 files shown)


client-side.md
```diff
@@ -29,23 +26,9 @@ free to justify why not.
...
-NPM, Bower, JSPM, etc? How to depend on other libraries?*
-
-Favour the following libraries unless you can justify otherwise:
-- [lodash](https://lodash.com/) over [underscore](http://underscorejs.org/)
-- [reqwest](https://github.com/ded/reqwest) for AJAX
-- [moment](http://momentjs.com/) for dates
-
-*TODO: any others?*
-
+- Do not commit your dependencies into Git. Instead, lock your dependencies (direct and transitive) using a lockfile.
```
SiAdcock, line 31: In some circumstances committing your dependencies can be helpful (e.g. committing the Yarn binary using [`yarn policies`](https://classic.yarnpkg.com/en/docs/cli/policies))

gtrufitt: Is that useful as a generalised recommendation or is it fairly specific to certain use-cases?

client-side.md
```diff
@@ -96,12 +74,6 @@ various areas below.
...
-
-*TODO: Would be good to have something about performance (memory,
-responsiveness, etc), maybe a strategy to assess performance, some
-targets, automated testing or RUM?*
+- Define web performance budgets and track them
 
 ### Build tools/task runners
-
-- Have an automated process to produce your final assets (e.g. using
-  Grunt, Gulp, [Plumber](https://github.com/plumberjs/plumber), etc).
```
SiAdcock, line 107: Why remove this one? The example tooling is a bit dated, admittedly.

client-side.md
SiAdcock: Yeah, it's very specific. Maybe we could soften to:

```suggestion
- Prefer locking your dependencies (direct and transitive) using a lockfile rather than committing them into Git.
```

client-side.md
```diff
@@ -96,12 +74,6 @@ various areas below.
 
 - Don’t fail silently.
 - Report your JavaScript errors into a logging system (e.g. Sentry).
-
-*TODO: Would be good to have something about performance (memory,
-responsiveness, etc), maybe a strategy to assess performance, some
-targets, automated testing or RUM?*
+- Define web performance budgets and track them
 
 ### Build tools/task runners
```
gtrufitt, line 79:
```suggestion
### Build tools/task runners

- Have an automated process to produce your final assets (e.g. webpack, rollup, microbundle ).
```

client-side.md
gtrufitt: 👍  Have updated with a more modern tool list.

PR #38, opened 2020-09-17
add `@guardian` packages recommendation

sndrs:
Here's the stuff @JamieB-gu and I put together for our `@guardian` npm packages recommendation – what do you think?

we had a further question:

- should we recommend publishing commonjs and esmodules, or just esm? @SiAdcock did you have to include cjs in source?

cc @guardian/client-side-infra  

_refs [this discussion](https://github.com/orgs/guardian/teams/client-side-infra/discussions/11) and https://github.com/guardian/dotcom-rendering/pull/1919_

### DIFF
--- a/client-side.md
+++ b/client-side.md
@@ -6,10 +6,11 @@ free to justify why not.
 
 ## `@guardian` packages
 ### Publishing
-- Publish packages in ES2020 JavaScript
+- Publish packages in ES2020 JavaScript (see `target` in [TypeScript's compiler options](https://www.typescriptlang.org/docs/handbook/compiler-options.html))
+- Do not ship or depend on polyfills
 - Prefer including TypeScript types (e.g. [declaration files](https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html))
 ### Consuming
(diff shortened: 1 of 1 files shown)


client-side.md
```diff
@@ -4,6 +4,15 @@ These are recommendations from the experiences of multiple teams
 across the Guardian. Feel free to not follow of them, but also feel
 free to justify why not.
 
+## `@guardian` packages
+### Publishing
+- Publish packages in ES2020 JavaScript
```
tjmw, line 9: Since we use TypeScript so frequently, do you think it would be worth mentioning or linking to _how_ to do this in that context? i.e. the `target` param in [TypeScript config options](https://www.typescriptlang.org/docs/handbook/compiler-options.html).

tjmw (approved):
Looks great! :+1:

Really good question about publishing esm or commonjs, it would be great to have that be part of the recommendations too (though no need to block this PR on figuring that out I don't think).

client-side.md
sndrs: yeah great idea, will add that

PR #42, opened 2020-11-19
[AWS Lambda] How to fix socket timeout issues

mbarton:
Add the workaround for socket issues with NodeJS lambdas. This isn't really a recommendation per-se but this repo is the best place I could think of to put the advice so it's not lost in chat rooms and emails


mchv:
At least it is documented somewhere we could reference.

PR #48, opened 2021-01-19
Add recommendations for continuous integration, delivery and deployment 

stephengeller:
## What does this change?

This PR adds documentation on continuous integration, deployment and delivery at the Guardian.

Following @guardian/developer-experience 's sessions on establishing a vision for CI & CD, these documents list the current state of play for these areas as well as our vision for what it could look like in the future.

For more information and context, please see the [Best Practices Index](https://docs.google.com/document/d/1CYftqM2xgLFu9yy2ZeZjLFufWbu_x7J8z3QhoJefA4o/edit#) created from the above sessions, with links to research and proposal documents inside.

## How to test

Verify that these documents align with our current practices and assess whether the proposal is appropriate for our department.

We have a clear set of documents listing our best practices for continuous integration, delivery and deployment.


davidfurey:
Thanks for contributing to the recommendations repo! There are definitely some great things in here.  I think it makes sense to mark this as a draft PR for now, and not consider merging it until it provides clear recommendations for the approach developers should use now.

Also, since this is a public repo ideally the final version would not contain links to docs that are not public.

stephengeller:
Thanks @davidfurey , that's good advice. I'll take those notes onboard - should be feasible to at least make all information public, and I can capture current approaches better here too.

stephengeller:
@davidfurey  your suggestions have been actioned and this is hopefully now ready for review, let me know if there is more needed.

davidfurey:
Thanks for the changes, it is much clearer now.  What do you think about moving the future vision into a github issue or Google doc?  My preference would be for the recommendations repo to be a living document of what we currently recommend rather than a place to articulate our vision.  But I'd like to hear your and others views, because maybe future vision does belong here.

stephengeller:
Interesting - I think you're right, having in one of those two places could be better as I do agree that we should keep this centred around recommendations rather than visions.

I'd worry that the vision would get lost if in a Google doc though, so maybe an issue or even a discussion would be more suitable? It would mean we have what we _are_ doing colocated with what we _want_ to be doing in the future. Also keen to hear other people's thoughts though.

stephengeller:
Following a conversation on chat with @davidfurey , we have agreed to move the proposals side of this PR to a GitHub discussion and keep this PR focused on current recommendations. Will update this PR accordingly soon!

stephengeller:
Future vision has been moved to a discussion (https://github.com/guardian/recommendations/discussions/50) and removed from this PR.

davidfurey (approved):

continuous-deployment.md
```diff
@@ -0,0 +1,20 @@
+Continuous Deployment
+-----
+
+Continuous Deployment (CD) is a software release process where verified changes to a codebase are deployed immediately and autonomously to a production environment.
+
+You should have continuous deployment set up for your repository, with the following behaviours configured if applicable:
+* Automatically deploying changes to the `main` branch to production
+* Regularly redeploying the application on a schedule
```
akash1810, line 8: IIUC scheduled deploys are mostly beneficial on projects using AMIgo images as it ensures the running instances are using recently baked AMIs.

Do we want to qualify this point at all?

```suggestion
* Using an AMIgo AMI and regularly redeploying the application on a schedule
```

stephengeller: I agree, that's an important thing to highlight. Updated with your suggestion

akash1810 (approved):

continuous-integration.md
```diff
@@ -0,0 +1,29 @@
...
+* Linting your codebase
+* Publishing artifacts
+* Other appropriate tasks for your repository
+
+## Platforms
+
+* Use TeamCity to run continuous integration tasks
+* If possible, have a centralised script in the repository named `scripts/ci`  
+    - This adheres to GitHub's [Scripts To Rule Them All pattern](https://github.com/github/scripts-to-rule-them-all)
+* For public repositories, you can use [GitHub Actions](https://docs.github.com/en/actions) for some above tasks, but we are yet to develop a safe way of uploading artifacts to RiffRaff through it
```
markjamesbutler, line 19: Should we be suggesting this if there isn't a safe way to do it? Probably we need to discuss what is mean't by safe.

stephengeller: Sorry, didn't see this comment until now! This is a good point, but I think this could be better explained by changing the phrasing a bit. 

I've just pushed a change that aims to rephrase the recommendation from "there isn't a safe way to push artefacts to RiffRaff" to "we haven't developed a best practice for this yet", which is more accurate. I think, at the time, I meant "safe" as "you can't safely use the recommended `node-riffraff-artifact` in GitHub Actions as  you'd have to publish/configure credentials somewhere", but this would be fairly safe to do if you configured AWS credentials in each repo and rotated them regularly.

I think a good  (and safe)solution for this is possible (likely relying on Guardian org-wide tokens being set up and a custom GitHub Action to implement it), and I think it's sensible to still encourage people to use GitHub Actions as it is a great interface, free for public repositories, and safe to use for the tasks mentioned besides publishing artefacts to RiffRaff.

continuous-integration.md
```diff
@@ -0,0 +1,31 @@
...
+* Running unit and integration tests
+* Building your application
+* Linting your codebase
+* Publishing artifacts
+* Other appropriate tasks for your repository
+
+## Platforms
+
+* Use TeamCity to run continuous integration tasks
+* If possible, have a centralised script in the repository named `scripts/ci`
```
akash1810, line 19: minor: Could we add the qualifier that CI should have a single step, and that is to run this script:

`Where possible, have CI execute a single, centralised script in the repository named scripts/ci`

akash1810 (approved):

continuous-deployment.md
```diff
@@ -0,0 +1,20 @@
+Continuous Deployment
+-----
```
mxdvl, line 2: Small markdown formatting issue, this results in `h2`.

```suggestion
Continuous Deployment
=====================
```

continuous-deployment.md
```diff
@@ -0,0 +1,20 @@
...
+
+Continuous Deployment (CD) is a software release process where verified changes to a codebase are deployed immediately and autonomously to a production environment.
+
+You should have continuous deployment set up for your repository, with the following behaviours configured if applicable:
+* Automatically deploying changes to the `main` branch to production
+* Using an AMIgo AMI and regularly redeploying the application on a schedule
+* Restricting deployments if and when required
+
+You should already have [continuous integration](continuous-integration.md) configured in order to set up continuous deployment.
```
mxdvl, line 11: A small Markdown note to say you can use the link text itself, and declare the link it should point to further down. It often reads nicer.

```suggestion
You should already have [continuous integration] configured in order to set up continuous deployment.

[continuous integration]: continuous-integration.md
```

continuous-deployment.md
```diff
@@ -0,0 +1,20 @@
...
+This ensures the deployed code is safe and ready for deployment. 
+
+## Platforms
+
+* Use [RiffRaff](https://github.com/guardian/riffraff) to deploy your application, set up scheduled deploys and configure continuous deployment
+    - Set up scheduled deploys, continuous deployment and deployment restrictions in the RiffRaff UI
+    - Configure your RiffRaff deployment types inside your repository's `riff-raff.yaml` file 
+
+**See [continuous integration](continuous-integration.md)**
```
mxdvl, line 20: Same as above, if you've added the link this will be turned into it automatically.

```suggestion
**See [continuous integration]**
```

continuous-integration.md
```diff
@@ -0,0 +1,31 @@
+Continuous Integration
+====
+
+Continuous Integration (CI) is a practice where developers frequently push their code into a shared repository. 
+Automated builds and checks run on each check-in, allowing teams to detect problems early.
+
+## Tasks
+
+You should have continuous integration set up in your repository running the following tasks where relevant:
+* Running unit and integration tests
```
mxdvl, line 10: Should we make a note of UI / visual regression testing here? The interface of our applications is often an important part of the integration process.

mxdvl (commented):
Mostly Markdown formatting comments from me.

I noticed we didn’t mention anything about visual testing or code coverage. Is it too team-specific to make it into a recommendation?

continuous-deployment.md
stephengeller: Oh, nice! Didn't know this is a thing, thanks for sharing it.

continuous-integration.md
stephengeller: Yeah, good call. Added this. Also corrected my own error of `scripts` -> `script` in the process

PR #49, opened 2021-01-20
Recommend np for publishing JS packages

tjmw:
Recommend [np](https://www.npmjs.com/package/np) for publishing JS packages.


oliverlloyd (approved):
Hearty 👍

client-side.md
```diff
@@ -5,15 +5,19 @@ across the Guardian. Feel free to not follow of them, but also feel
 free to justify why not.
 
 ## `@guardian` packages
+
 ### Publishing
+
 - Publish packages in ES2020 JavaScript (see `target` in [TypeScript's compiler options](https://www.typescriptlang.org/docs/handbook/compiler-options.html))
 - Do not ship or depend on polyfills
 - Prefer including TypeScript types (e.g. [declaration files](https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html))
+- Use [np](https://www.npmjs.com/package/np) to publish packages
```
sndrs, line 14: would it be worth suggesting adding it via a release npm-script? i.e. not just a global install

tjmw: Good question! That's the route I took when I started using `np` recently, so suggesting that seems fine to me. I know @oliverlloyd mentioned being happy about _not_ having to add it as a dependency to the project and just using a global install so would be interested in his thoughts.

Alternatively maybe there's another separate recommendation we could make about preferring to add any tooling required for a project as a direct (dev) dependency, rather than relying on global installs? It feels like this might be more generally applicable than just to `np`. What do you reckon?

oliverlloyd: I think @sndrs is right that it, ideally, would be a dependency and in general I prefer project level installs of things like this. It's useful to point out though that `np` just works by itself and isn't _needed_ in package.json

tjmw: Cool, thanks. I've added some wording to include @sndrs's suggestion.

oliverlloyd: That new wording makes sense 👍 

I wonder if it's worth saying something about how to ensure the project is built? I ask for selfish reasons because I'm actually curious to learn more about this! Because we have projects using both `npm` and `yarn` I'm not sure when I can be confident the build will be a pre release step and not

oliverlloyd: Note. I'm happy for this to be a separate PR so we can unblock this one.

tjmw: Thanks @oliverlloyd. Good point about ensuring the project is built. Agreed though that separate PR makes sense 👍

PR #52, opened 2021-02-24
Do not not publish typescript

sndrs:
update npm package recommendations to never publish typescript source files

### DIFF
--- a/client-side.md
+++ b/client-side.md
@@ -11,7 +11,7 @@ free to justify why not.
 - Publish packages in ES2020 JavaScript (see `target` in [TypeScript's compiler options](https://www.typescriptlang.org/docs/handbook/compiler-options.html))
 - Do not ship or depend on polyfills
 - Include TypeScript [declaration files](https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html) if your project uses TypeScript.
-  - Never publish `.ts`/`.tsx` files – they depend on an unknownable, per-project TypeScript configuration.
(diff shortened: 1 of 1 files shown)


client-side.md
```diff
@@ -10,7 +10,8 @@ free to justify why not.
 
 - Publish packages in ES2020 JavaScript (see `target` in [TypeScript's compiler options](https://www.typescriptlang.org/docs/handbook/compiler-options.html))
 - Do not ship or depend on polyfills
-- Prefer including TypeScript types (e.g. [declaration files](https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html))
+- Include TypeScript [declaration files](https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html) if your project uses TypeScript.
+  - Never publish `.ts`/`.tsx` files – they depend on an unknownable, per-project TypeScript configuration.
```
tjmw, line 14: Super nit-picky language suggestion:

```suggestion
  - Do not publish `.ts`/`.tsx` files – they depend on an unknownable, per-project TypeScript configuration.
```

Just because that's the phrasing we've used above for polyfills, and as it is I might interpret this `never` as somehow being more important than `do not`, which I don't think is the case? What do you reckon?

tjmw (approved):

client-side.md
sndrs: agree

SiAdcock (approved):

PR #53, opened 2021-03-01
Create separate npm-packages doc

sndrs:
## What does this change?
creates a separate npm-packages doc and links back to it from the client side one.

### DIFF
--- a/npm-packages.md
+++ b/npm-packages.md
@@ -1,0 +1,46 @@
+# Publishing and using `@guardian` NPM packages
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Using
+
+-   Make sure you include any `@guardian` packages for transpilation in your build process
+    -   e.g. your [`webpack.config.js`](https://github.com/webpack/webpack/issues/2031#issuecomment-219040479) might including something that looks like this:
+
+```js
+module: {
+	rules: [
+		{
+			test: /\.m?(j|t)sx?$/,
+			use: {
+				loader: 'babel-loader',
+			},
+			exclude: {
+				test: /node_modules/,
+
+				// don't exclude '@guardian' node_modules
+				exclude: /@guardian\//,
+			},
+		},
+	]
+}
+```
+
+## Publishing
+
+### Use `@guardian`
+
+-   Publish to NPM using the [`@guardian`](https://www.npmjs.com/org/guardian) scope
+
+### Package content
+
+-   Publish packages in ES2020 JavaScript
+    -   For TypeScript projects, see `target` in the [compiler options](https://www.typescriptlang.org/docs/handbook/compiler-options.html)
+-   Do not ship or depend on polyfills
+
+#### TypeScript
+
+-   Create and include [declaration files](https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html) if your project uses TypeScript
+-   Do not publish `.ts`/`.tsx` files
+    -   They depend on a project-specific TypeScript config, and your config may not be the same the project that installs your package
--- a/client-side.md
+++ b/client-side.md
@@ -4,44 +4,33 @@ These are recommendations from the experiences of multiple teams
 across the Guardian. Feel free to not follow of them, but also feel
 free to justify why not.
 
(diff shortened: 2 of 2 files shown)


gtrufitt (approved):

npm-packages.md
```diff
@@ -0,0 +1,46 @@
+# Publishing and using `@guardian` NPM packages
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Using
+
+-   Make sure you include any `@guardian` packages for transpilation in your build process
+    -   e.g. your [`webpack.config.js`](https://github.com/webpack/webpack/issues/2031#issuecomment-219040479) might including something that looks like this:
```
oliverlloyd, line 9: Could we also add examples for tooling such as storybook and jest? Maybe we examples of the sorts of errors that can be expected to be seem if this isn't done?

I've often hit this issue in multiple reports for multiple tools and even though I've fixed it several times it's easy to forget to tell the config to not exclude guardian packages and be faced with confusing error messages. I can imagine other people could have similar problems.

I think the motivation and reasoning behind this appraoch is sound and agree that we should do it but it is, nonetheless, counter intuative. We're all so used to npm packages being compiled down to something we don't need to worry about; we have all been blithly ignoring `node_modules` for so long it's become a given, we have coding muscle memory for it.

sndrs: > Could we also add examples for tooling such as storybook and jest?

yeah great idea, you ok to raise a pr?

>we have all been blithly ignoring node_modules for so long it's become a given, we have coding muscle memory for it

this was brought up in infra meetings actually, and we agreed we should really be transpiling everything, beucase you have no control over what's published by other people. but it's not formalised here – it probably should be

npm-packages.md
```diff
@@ -0,0 +1,46 @@
...
+		},
+	]
+}
+```
+
+## Publishing
+
+### Use `@guardian`
+
+-   Publish to NPM using the [`@guardian`](https://www.npmjs.com/org/guardian) scope
```
JamieB-gu, line 34: Have we decided that `npm` is the place to standardise, and we should migrate away from any attempts to use GitHub URLs or GitHub Packages?

If so, do we have any more specific recommendations for _how_ to publish to `npm`? E.g. using `np`, publishing from `main` etc.?

JamieB-gu (commented):
Looks very sensible!

npm-packages.md
gtrufitt: > Have we decided that npm is the place to standardise, and we should migrate away from any attempts to use GitHub URLs or GitHub Packages?

I'm not convinced on the value vs confusion this would cause? I'd rather the consistency and the effort to switch *all* to github or github packages - does it seem worth it?

tjmw: > If so, do we have any more specific recommendations for how to publish to npm? E.g. using np, publishing from main etc.?

Oh, I just noticed that in moving to a separate file we lost a recommendation about using `np` to publish. @sndrs I'm not sure if that was intentional? We used to have this:

```
Use [np](https://www.npmjs.com/package/np) to publish packages. Add it as a project dependency and call it via a `release` script in your `package.json`.
```

JamieB-gu: > I'd rather the consistency and the effort to switch _all_ to github or github packages - does it seem worth it?

On *GitHub_ dependencies*, this means using the [`github:` prefix and `#<commit-ish>`](https://docs.npmjs.com/cli/v6/commands/npm-install) to specify a repo and branch/commit/tag. This comes with advantages and disadvantages. We're using it [on AR](https://github.com/guardian/apps-rendering/blob/bd7214635ea43813673e2fb40db783e45069bd58/package.json#L45) for a few things - other projects like DCR would have to decide if that would work for them as well.

On *GitHub packages*, perhaps we should do another spike to see if they're easier to use since @oliverlloyd last looked into them? AFAIK there are some nice upsides (GitHub auth so no need for a separate npm account, ability to publish via GitHub Actions, etc.), but some awkward downsides (auth required to install packages?).

sndrs: > Oh, I just noticed that in moving to a separate file we lost a recommendation about using np to publish. @sndrs I'm not sure if that was intentional?

no an accident, will restore it

PR #55, opened 2021-03-17
Fill out npm package recommendations

sndrs:
## What does this change?

Add more details and further recommendations for creating and using `@guardian` NPM packages

### DIFF
--- a/npm-packages.md
+++ b/npm-packages.md
@@ -3,7 +3,7 @@
 <!-- START doctoc generated TOC please keep comment here to allow auto update -->
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
-## Publishing a `@guardian` package to NPM
+## Publishing a `@guardian` library as an NPM package
 
 ### Authoring
 
@@ -11,15 +11,15 @@ Write your library in [TypeScript](https://www.typescriptlang.org).
 
 ### Packaging
 
-Do not publish your source code in your package, especially if you use TypeScript.
+Do not include your TypeScript source code in your package. Instead, publish compiled JavaScript files and [declaration files](https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html).
 
 > TypeScript files depend on a project-specific configuration, and your setup may not be the same the project that installs your package
 
-Do not include or depend on [polyfills](https://developer.mozilla.org/en-US/docs/Glossary/Polyfill).
+Do not include or depend on [polyfills](https://developer.mozilla.org/en-US/docs/Glossary/Polyfill). Leave these decisions to the projects that install your package (see the section on using our packages below).
 
 #### Compiling
 
-Use the TypeScript compiler (`tsc`) to generate the JavaScript files and TypeScript [declaration files](https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html) that you will publish.
+Use the TypeScript compiler (`tsc`) to generate the JavaScript files and TypeScript [declaration files](https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html) that you publish.
 
 Generate two versions of your library:
 
--- a/npm-packages.md
+++ b/npm-packages.md
@@ -11,7 +11,7 @@ Write your library in [TypeScript](https://www.typescriptlang.org).
 
 ### Packaging
 
-Do not include your TypeScript source code in your package. Instead, publish compiled JavaScript and [declaration files](https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html).
+Do not include your TypeScript source code in your package. Instead, compile your source to JavaScript and [declaration files](https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html) and publish those.
 
 > TypeScript depends on a project-specific configuration, and your setup may not be the same the project that installs your package
 
--- a/npm-packages.md
+++ b/npm-packages.md
@@ -11,22 +11,22 @@ Write your library in [TypeScript](https://www.typescriptlang.org).
 
 ### Packaging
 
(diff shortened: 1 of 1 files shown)


faresite:
Great recommendations @sndrs , I was just going back and forth a moment ago thinking where I should build the types in my atoms-rendering update .. and even a small recommendation like this already takes away some of that unnecessary decision making 👍

npm-packages.md
```diff
@@ -1,14 +1,115 @@
...
+
+Use [Semantic Release](https://github.com/semantic-release/semantic-release) in a GitHub action with [guardian/actions-merge-release-changes-to-protected-branch](https://github.com/guardian/actions-merge-release-changes-to-protected-branch) to automate publishing.
+
+#### Spontaneous publishing
+
+Publish manually from the command line using [np](https://www.npmjs.com/package/np).
+
+## Using `@guardian` NPM packages
+
+To ensure the Guardian's application bundles are as efficient as possible, packages assume nothing about where they will run.
```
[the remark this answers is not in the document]
sndrs: that's correct, you can declare a dependency like that, but this refers the environment (browser version etc). i'll try to make to make that clearer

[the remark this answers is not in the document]
faresite: Got it 👍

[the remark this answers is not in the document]
faresite: That leads me to suggesting adding some recommendations in the **Authoring** section about what to include in `dependencies` vs `peerDependencies`?
One of the problems I recently had was caused by `image-rendering` declaring `@guardian/types` both in `dependencies` and `peerDependencies` with a minor version off. The type errors I was getting from TypeScript by no means were helpful and took me a while to notice that the problem wasn't in the `apps-rendering` codebase!
Just having a quick look at `discussion-rendering` and I can see that there's a babel plugin in its `dependencies` but this seems like a blip.
In `atoms-rendering` there's only a youtube package in its `dependencies` and all the `@guardian` packages live in `peerDependencies` which seems to me like a good call. 
Just mentioning it because if I were to author a new package that depends on `@guardian` packages I'd probably just copy the config from another repo without thinking too much about what package goes where (I probably shouldn't admit to that 🙈).

jamie-lynch (approved):
👍  This looks good to me. As we said yesterday, maybe we can move some of the information from the [actions-merge-release-changes-to-protected-branch README](https://github.com/guardian/actions-merge-release-changes-to-protected-branch/blob/main/README.md) into here (or at least somewhere in this repo) at some point but happy for this to get merged as is for now.

npm-packages.md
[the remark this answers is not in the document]
sndrs: i'm not sure yet _how_ to handle these properly. there seems good arguements to use `peerDependencies` and normal `dependencies` so will leave this for now and come back to it (promise)

npm-packages.md
```diff
@@ -1,14 +1,117 @@
...
+
+Write your library in [TypeScript](https://www.typescriptlang.org).
+
+### Packaging
+
+Do not include your TypeScript source code in your package. Instead, compile your source to JavaScript and [declaration files](https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html) and publish those.
+
+> TypeScript depends on a project-specific configuration, and your setup may not be the same the project that installs your package
+
+Do not include or depend on [polyfills](https://developer.mozilla.org/en-US/docs/Glossary/Polyfill). Let projects that install your package decide what they need (see the section on using our packages below).
```
SiAdcock, line 18: I wonder if this needs fleshing out or defining further. I guess whether a polyfill is needed depends on the target browsers that a project supports. So we're rightly saying that projects, not libraries, should decide which polyfills to bring.

But what does it mean to say to library authors "Do not depend on polyfills"? Are we saying "Do not depend on language features that require polyfills as of ES2020?"

Let me know if what I'm saying makes no sense!

npm-packages.md
```diff
@@ -1,14 +1,117 @@
...
+
+#### Compiling
+
+Use the TypeScript compiler (`tsc`) to generate the JavaScript and TypeScript [declaration files](https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html) that you publish.
```
SiAdcock, line 22: Do we want to drop a line in here saying why we don't recommend `@babel/preset-typescript`? (it doesn't support all TS language features)

npm-packages.md
```diff
@@ -1,14 +1,117 @@
...
+
+Generate two versions of your library:
+
+1. ES modules targetting ES2020
+2. CommonJS modules targetting ES2018 (Node 10+)
```
SiAdcock, line 27: Out of interest, where did you find out what version of ECMAScript Node.js 10 supports?

Also maybe we should say Node 12+ here since we're about to kick everyone off of Node 10 😄

npm-packages.md
sndrs: maybe we should have separate typescript recommendations for that? maybe in the general client side ones, or pulled out?

npm-packages.md
sndrs: i got it from SO :) but also https://node.green/#ES2018 

if we go with 12 we can target 2019

npm-packages.md
sndrs: ah right, no i meant to say that you can write anything that TS will let you when you target ES2020, and applications will manage making the code you export work.

does it need rewording?

npm-packages.md
SiAdcock: Wow I have never scrolled down that far! Thanks for the tip! 🎩

npm-packages.md
SiAdcock: Yeah, I guess it does feel bigger than a library recommendation. General client side makes sense.

npm-packages.md
SiAdcock: I think so? I'll have a go at it and see what people think

PR #58, opened 2021-04-20
Don't delete S3 buckets

mbarton:
Recommendation to not delete S3 buckets along with the rationale behind it


jonathonherbert (approved):
Clear to me 👍

PR #64, opened 2021-05-26
--cli-read-timeout to prevent AWS CLI running lambdas multiple times

joelochlann:
Add recommendations on using the `--cli-read-timeout` param for the CLI to prevent unintended multiple lambda invocations

### DIFF
--- a/AWS.md
+++ b/AWS.md
@@ -42,6 +42,7 @@ S3
 Lambda
 ------
 
+## Unexpected socket errors in Node v10 and 12 lambdas
 You may occasionally see unexpected socket errors in NodeJS lambdas:
 
 ```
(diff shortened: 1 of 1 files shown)


shtukas (approved):

AWS.md
```diff
@@ -87,6 +88,19 @@ export function handler(event: Event, context: any, callback: (err: Error | null
 
 AWS advise you should only use the workaround if you see errors.
 
+### Long-running lambda triggered multiple times from CLI `aws lambda invoke`
+
+The AWS CLI has a timeout set by [`--cli-read-timeout`](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-options.html#:~:text=cli%2Dread%2Dtimeout), which determines how long (in seconds) it will wait for data on an open socket.
+
+The implicit default seems to be 60 seconds. If you use the CLI to trigger a lambda synchronously (`--invocation-type RequestResponse`), and your if the lambda runs longer than 60 seconds, the CLI will terminate the first request and trigger a retry. Your lambda will execute multiple times, despite the first invocation having succeeded. (Question: is this retry caused by the lambda service, or the CLI?).
```
mbarton, line 95:
```suggestion
The implicit default seems to be 60 seconds. If you use the CLI to trigger a lambda synchronously (`--invocation-type RequestResponse`) and the lambda runs longer than 60 seconds, the CLI will abandon the first request and trigger a retry
```

mbarton: Minor wording suggestion. I think there's also a `your if` typo

AWS.md
```diff
@@ -87,6 +88,19 @@ export function handler(event: Event, context: any, callback: (err: Error | null
...
+
+The solution is to set `--cli-read-timeout` to a value longer than your lambda timeout, or set it to `0` to remove the timeout and make the socket read wait indefinitely.
```
mbarton, line 97: I'm tempted to suggest `0` in all cases in case the lambda timeout gets changed and we forget to update this parameter in the script

joelochlann: yeah that's a great shout. always good to have a single recommendation rather than options

mbarton (approved):

shtukas (approved):

PR #69, opened 2021-07-29
Add repository topics recommendation

jamie-lynch:
## What does this change?

This PR adds a new recommendation to `github.md` around categorising repositories using [topics](https://docs.github.com/en/github/administering-a-repository/managing-repository-settings/classifying-your-repository-with-topics). This makes it clear to others what the purpose of the repository is and allow projects to be filtered out of searches where required.

### DIFF
--- a/github.md
+++ b/github.md
@@ -19,3 +19,18 @@ It is recommended to include a `CODEOWNERS` file which references the GitHub tea
 [GitHub environments](https://docs.github.com/en/actions/reference/environments) can be configured with protection rules and secrets. Workflows that reference an environment will then be bound by its protection rules and, if allowed, get access to it's secrets. When configuring workflows that require secrets, consider whether an environment can be used in conjunction with the [branch protection rules](#branch-protection) to limit the use of the secret.
 
 For example, when configuring [auto-publishing for an npm library](./npm-packages.md#continuous-delivery) it is recommended that you create a new environment for this purpose, containing the NPM and GitHub tokens and with the protection rules set so that it can only be used from the release branch. The release workflow should then reference this environment.
+
+## Topics
+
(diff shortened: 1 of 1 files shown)


github.md
```diff
@@ -19,3 +19,18 @@ It is recommended to include a `CODEOWNERS` file which references the GitHub tea
...
+## Topics
+
+[Topics](https://docs.github.com/en/github/administering-a-repository/managing-repository-settings/classifying-your-repository-with-topics) should be used to categorise repositories that are not destined for production. These topics make it clear to others what the purpose of the repository is and allow projects to be filtered out of searches where required. 
+
+|   Topic   | Description |
+| --------- | ----------- | 
+| prototype | Repositories that are used to prove a new concept * |
+| learning  | Repositories that have been created as part of a learning exercise |
+| research  | Repositories that have been created as part of researching or trialing a new language, technology or framework |
+| hackday   | Repositories that are created during hack days * |
```
jamie-lynch, line 32: Keen to hear suggestions on what topics we should have.

mchv: I am not sure we need `research` as I think `prototype` and `learning` are enough.
I ❤️ `hackday` as it is a great way to find things we have created or originally comes from a hackday.

mchv: i.e I would add `hackday` to `ophan` repository for instance

jamie-lynch: That's true, it would be really nice to have that! One use case for these topics would be for filtering out of searches any repositories that aren't production but also that we don't want to archive (e.g. when DevX are looking for deprecated patterns) so maybe we need a `production` topic too?

mchv:
Let’s try and see how it goes!
@jamie-lynch may be you could start by tagging few projects to raise awareness?

PR #72, opened 2021-10-19
Add new GH topics

jamie-lynch:
## What does this change?

This change adds two new topics to the list of recommendations for tagging GitHub repos. 

The `testing` topic is added as the existing ones didn't sufficiently cover the use case of some repositories. 

The `production` topic has been added so that other topics (e.g. `hackday`) can be maintained when a repository is productionised.

### DIFF
--- a/github.md
+++ b/github.md
@@ -28,12 +28,12 @@ For example, when configuring [auto-publishing for an npm library](./npm-package
 
 [Topics](https://docs.github.com/en/github/administering-a-repository/managing-repository-settings/classifying-your-repository-with-topics) should be used to categorise repositories that are not destined for production. These topics make it clear to others what the purpose of the repository is and allow projects to be filtered out of searches where required.
 
-| Topic     | Description                                                        |
-| --------- | ------------------------------------------------------------------ |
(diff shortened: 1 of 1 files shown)


davidfurey:
Is the intention that list becomes becomes mutual exclusive, collectively exhaustive?  If so, perhaps we should namespace these topics in some way.

jamie-lynch:
> Is the intention that list becomes becomes mutual exclusive, collectively exhaustive? If so, perhaps we should namespace collectively exhaustive

I don't think these topics should all be mutually exclusive because, as @mchv pointed out, it would be nice to have the `hackday` label applied to all projects that originated from hackdays, including those that go on to make it to production. 

Given that, I don't think they can be collectively exhaustive without some sort of grouping or namespacing so that sounds like a good idea. 

It'd be good to have some wider input into what those namespaces might be and what other use cases teams might have for them before we start applying them too widely and then having to go back through and make changes so I'm going to create a new discussion and collate all of the context and conversations to date there.

mchv:
I am fine for merging this and evolving this when we find issues or need better categorisation

PR #73, opened 2021-12-31
Add Emotion recommendations

SiAdcock:
## What does this change?

[Emotion](https://emotion.sh) is integral to how we implement CSS at the Guardian. However it is complex, and it can be hard to configure. This change provides a set of recommendations for configuration depending on the tech choices of the team using it.

This documentation was previously in the Source component library docs simply because Source requires consumers to also use Emotion. It makes more sense as a set of general recommendations as it doesn't specifically relate to the Source component library.

### DIFF
--- a/emotion.md
+++ b/emotion.md
@@ -1,0 +1,81 @@
+# Using Emotion
+
+[Emotion](https://emotion.sh/) is a CSS-in-JS library that provides component-scoping, encapsulation and composition for your styles.
+
+## Getting started
+
+```shell
+$ yarn add @emotion/react
+```
+
+> You need to install `@emotion/react` to use [Source components](https://github.com/guardian/source)
+
+## Configuration
+
+### TypeScript
+
+We recommend using [React's automatic runtime](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html).
+
+In your `tsconfig.json`:
+
+```json
+{
+  "compilerOptions": {
+    "jsx": "react-jsx",
+    "jsxImportSource": "@emotion/react"
+  }
+}
+```
+
+> This tells TypeScript to use React's automatic runtime and lets TypeScript know where it can find the JSX namespace.
+
+### Babel
+
+You can use the `css` prop, in conjunction with `@emotion/babel-preset-css-prop`:
+
+```shell
+$ yarn add @emotion/babel-preset-css-prop -D
+```
+
+In your `.babelrc`:
+
+```json
+{
+  "presets": ["@emotion/babel-preset-css-prop"]
+}
+```
+
+> **Note:** If you use `@babel/preset-react` or `@babel/preset-typescript` ensure that `@emotion/babel-preset-css-prop` is inserted _after_ them in your babel config.
+
+### Server side rendering
+
+If your app is doing server-side rendering, you need to follow the [Advanced Approach](https://emotion.sh/docs/ssr#advanced-approach) to SSR setup. This is because our components use sibling selectors that may be interrupted by the way Emotion embeds `<style>` tags directly into your markup.
+
+## Example
+
+```jsx
+import { css } from "@emotion/react"
+
+const color = "darkgreen"
+
+render(
+  <div
+    css={css`
+      background-color: hotpink;
+      &:hover {
+        color: ${color};
+      }
+    `}
+  >
+    This has a hotpink background.
+  </div>
+)
+```
+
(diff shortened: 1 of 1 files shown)


emotion.md
```diff
@@ -0,0 +1,81 @@
+# Using Emotion
```
sndrs, line 1:
```suggestion
# Using Emotion with React
```

we're specifically setting this is up for use with react right?

emotion.md
```diff
@@ -0,0 +1,81 @@
...
+
+[Emotion](https://emotion.sh/) is a CSS-in-JS library that provides component-scoping, encapsulation and composition for your styles.
```
sndrs, line 3:
```suggestion
[Emotion](https://emotion.sh/) is a CSS-in-JS library that provides component-scoping, encapsulation and composition for your styles.

It can be used with and without React, but this guide covers our standard use, which is alongside React. In particular it sets out how to use Emotion alongside the React implementation of our design system, [`Source`](https://github.com/guardian/source).
```

emotion.md
```diff
@@ -0,0 +1,81 @@
...
+    "jsxImportSource": "@emotion/react"
+  }
+}
+```
+
+> This tells TypeScript to use React's automatic runtime and lets TypeScript know where it can find the JSX namespace.
+
+### Babel
+
+You can use the `css` prop, in conjunction with `@emotion/babel-preset-css-prop`:
```
sndrs, line 34: do you need this if you have the above ts setup?

emotion.md
```diff
@@ -0,0 +1,81 @@
...
+{
+  "presets": ["@emotion/babel-preset-css-prop"]
+}
+```
+
+> **Note:** If you use `@babel/preset-react` or `@babel/preset-typescript` ensure that `@emotion/babel-preset-css-prop` is inserted _after_ them in your babel config.
+
+### Server side rendering
+
+If your app is doing server-side rendering, you need to follow the [Advanced Approach](https://emotion.sh/docs/ssr#advanced-approach) to SSR setup. This is because our components use sibling selectors that may be interrupted by the way Emotion embeds `<style>` tags directly into your markup.
```
sndrs, line 52:
```suggestion
If your app is doing server-side rendering, you need to follow [Emotion's Advanced Approach](https://emotion.sh/docs/ssr#advanced-approach) to SSR setup. This is because our components use sibling selectors that may be interrupted by the way Emotion embeds `<style>` tags directly into your markup.
```

emotion.md
SiAdcock: No, I believe this is handled by the [`jsxImportSource` option](https://emotion.sh/docs/typescript#emotionreact) in your tsconfig's `compilerOptions`:

```
{
  "jsxImportSource": "@emotion/react"
}
```

SiAdcock: Should we make it clearer that TypeScript and Babel are 2 different paths for configuration? e.g

```
## Configuration

Emotion may be configured in your `tsconfig.json` or your `.babelrc`, depending on your build setup.

### Approach 1: TypeScript

...

### Approach 2: Babel

...
```

sndrs: 👍

emotion.md
```diff
@@ -0,0 +1,83 @@
...
+## Getting started
+
+```shell
+$ yarn add @emotion/react
+```
+
+> You need to install `@emotion/react` to use [Source components](https://github.com/guardian/source)
+
+## Configuration
+
```
sndrs, line 16:
```suggestion
## Configuration

Emotion can be configured for typescript or babel, depending on your setup. If you are using both, you only need to do one (TypeScript is simplest).
```

emotion.md
sndrs: added a suggestion so will resolve this

emotion.md
```diff
@@ -0,0 +1,83 @@
...
+### TypeScript
+
+We recommend using [React's automatic runtime](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html).
```
sndrs, line 19:
```suggestion
If you are using TypeScript <= 4.1 and React <=17, it's simplest to use [React's automatic runtime](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) with TypeScript's attendant [jsx](https://www.typescriptlang.org/tsconfig#jsx) and [jsxImportSource}(https://www.typescriptlang.org/tsconfig#jsxImportSource) compiler options.
```

PR #76, opened 2022-01-27
Add more detail AWS CFN recommendations

akash1810:
Recommending:
  - CDK
  - Tagging
  - Resource import
  - CD
  - One S3 bucket per stage
  - Private S3 buckets by default

Changes can be viewed [here](https://github.com/guardian/recommendations/blob/aa-aws-recommendations/AWS.md).

### DIFF
--- a/AWS.md
+++ b/AWS.md
@@ -1,7 +1,21 @@
 General
 -------
 
- * Use CloudFormation or ElasticBeanStalk to define infrastructure in AWS.
+ * Provision and manage all AWS resources using infrastructure as code
+   * Prefer to use [CDK](https://github.com/guardian/cdk) to generate CloudFormation. You might find that older projects still use CloudFormation directly; these should be migrated to CDK where possible.
+   * Prefer to [import resources](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-import.html) into CFN over manually configuring them
+ * Tag resources with:
+   * `Stack` - the broad umbrella the service sits under e.g. `media-service`
+   * `Stage` - the environment, typical values are:
+     * `PROD` for production
+     * `CODE` for pre-production
+     * `INFRA` for infrastructure or singleton resources e.g. [elasticsearch-node-rotation](https://github.com/guardian/elasticsearch-node-rotation)
+   * `App` - the individual service e.g. `image-loader`
+ * If a resource needs to be shared across multiple environments, prefer to define it in it's own CFN template as the same resource cannot be defined in multiple templates
+ * Prefer continuous delivery of infrastructure via Riff-Raff over manual deployment
+   * This provides a better audit trail
+   * This decreases risk as, by default, Riff-Raff will protect stateful resources from being accidentally deleted 
+ * Do not have private or secret information in a CFN template. Prefer to use Parameters or [private-infrastructure-config](https://github.com/guardian/private-infrastructure-config)
 
 Permissions
 -----------
@@ -34,6 +48,9 @@ SQS
 S3
 --
 
+ * Create buckets with CloudFormation, one per stage
+ * Block public access
+ * If public access is needed, enable this via a bucket policy rather than per object
  * Don't delete S3 buckets once they're not used any more
    * S3 bucket names are global and someone else could pick up the same name
    * This may pose a security risk if we inadvertently still use or reference the bucket


akash1810:
cc @guardian/developer-experience

AWS.md
```diff
@@ -1,7 +1,18 @@
 General
 -------
 
- * Use CloudFormation or ElasticBeanStalk to define infrastructure in AWS.
+ * Use [CDK](https://github.com/guardian/cdk) CloudFormation or ElasticBeanStalk to define infrastructure in AWS
```
jacobwinch, line 4: I find this first line a bit confusing. How about (something like):

> * Provision and manage all AWS resources using infrastructure as code
>   * Prefer to use [CDK](https://github.com/guardian/cdk) to generate CloudFormation where possible. You might find that older projects still use CloudFormation directly, these should be migrated to CDK where possible.

I also think we can remove any reference to `ElasticBeanStalk` here?

AWS.md
```diff
@@ -1,7 +1,18 @@
...
+   * Prefer to [import resources](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-import.html) into CFN over manually configuring them
+ * Tag resources with:
+   * `Stack` - the broad umbrella the service sits under e.g. `deployTools`
+   * `Stage` - the environment, typical values are:
+     * `PROD` for production
+     * `CODE` for pre-production
+     * `INFRA` for infrastructure or singleton resources e.g. [elasticsearch-node-rotation](https://github.com/guardian/elasticsearch-node-rotation)
+   * `App` - the individual service e.g. `prism`
+ * If a resource needs to be shared across multiple stacks, prefer to define it in it's own CFN template with stage `INFRA`
+ * Prefer continuous delivery of infrastructure via Riff-Raff over manual deployment
```
jacobwinch, line 14: It could be worth explaining the rationale (e.g. better audit trail and safety mechanisms in Riff-Raff) for this?

AWS.md
```diff
@@ -1,7 +1,18 @@
 General
 -------
 
- * Use CloudFormation or ElasticBeanStalk to define infrastructure in AWS.
+ * Use [CDK](https://github.com/guardian/cdk) CloudFormation or ElasticBeanStalk to define infrastructure in AWS
+   * Prefer to [import resources](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-import.html) into CFN over manually configuring them
+ * Tag resources with:
+   * `Stack` - the broad umbrella the service sits under e.g. `deployTools`
```
jacobwinch, line 7: Minor: we tend to use `deploy` as the stack name

AWS.md
akash1810: Yes, that's much better! Done.

AWS.md
akash1810: 👍🏽  Done.

AWS.md
akash1810: Ah yes. I might choose a different example as a `deploy` stack isn't particularly obvious...

AWS.md
```diff
@@ -1,7 +1,21 @@
...
+   * Prefer to use [CDK](https://github.com/guardian/cdk) to generate CloudFormation. You might find that older projects still use CloudFormation directly; these should be migrated to CDK where possible.
+   * Prefer to [import resources](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-import.html) into CFN over manually configuring them
+ * Tag resources with:
+   * `Stack` - the broad umbrella the service sits under e.g. `deployTools`
+   * `Stage` - the environment, typical values are:
+     * `PROD` for production
+     * `CODE` for pre-production
+     * `INFRA` for infrastructure or singleton resources e.g. [elasticsearch-node-rotation](https://github.com/guardian/elasticsearch-node-rotation)
+   * `App` - the individual service e.g. `prism`
+ * If a resource needs to be shared across multiple stacks, prefer to define it in it's own CFN template with stage `INFRA`
```
jacobwinch, line 14: I worry that this line might confuse people. For example, if there is a stream or queue which links together two accounts or stacks, it probably makes sense for this to have a `CODE` and `PROD` version, rather than a single `INFRA` stack. 

That said, I'm struggling to come up with a clear statement which explains when to use `INFRA` - I would hope that DevX are the only people who need to use this stage. We often seem to use it when it isn't possible (or doesn't quite make sense) to have a pre-production environment!

akash1810: I agree. What about dropping the bit about stage entirely?

```
If a resource needs to be shared across multiple environments, prefer to define it in it's own CFN template for clarity as a resource cannot be defined in multiple templates.
```

AWS.md
jacobwinch: Giving a good example here is tricky... 

I think 'Stack' provides the most benefits when we use it to mean _a logical grouping of microservices that are responsible for some particular functionality_ e.g. `mobile-notifications`. For example, this allows you to easily filter all logs responsible for sending a notification, regardless of which app/lambda is doing the work. But in practice for most teams `Stack` is synonymous with _the account that you want to deploy things into_.

akash1810: Yeah. I was hoping "umbrella" was generic/non-specific enough to cater for both of these use-cases. But maybe that's not helpful?

We've discussed before, but `Stack` doesn't really make sense anymore for the reasons you describe and also because it's overloaded.

AWS.md
jacobwinch: Yes, that looks good to me 👍

AWS.md
jacobwinch: I think the updated example makes this as clear as possible 👍  (I'd like to remove the concept of `Stack` entirely but that's slightly outside the scope of this PR 🙃 )

jacobwinch (approved):

PR #78, opened 2022-02-17
Update GitHub recommendations

akash1810:
## What does this change?

- Recommendations for the default branch name
- Recommendations for what to commit to a repository
- Recommendations on access

See the changes [here](https://github.com/guardian/recommendations/blob/aa-github/github.md).

### DIFF
--- a/github.md
+++ b/github.md
@@ -1,17 +1,15 @@
 # GitHub
 
 ## Repository Configuration
-
 ### Default branch name
-
 Use `main`.
 
-The [`master-to-main` tool][master-to-main] can be used to rename the default branch.
+Words matter. See docs from the [`master-to-main` tool][master-to-main] for more information. This tool can also be used to rename the default branch.
 
 ### Branch Protection
+Enable branch protection for the default branch.
 
-Branch protection should be enabled for the default/release branch of a repository, particularly when continuous delivery is configured. The following settings are recommended:
-
+Particularly when [continuous delivery] is configured, branch protection reduces risk as it means changes get reviewed before being deployed.  The following settings are recommended:
 - Require pull request reviews before merging
 - Require review from Code Owners (see [Codeowners](#codeowners))
 - Require status checks to pass before merging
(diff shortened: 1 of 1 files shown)


github.md
```diff
@@ -37,3 +46,50 @@ For example, when configuring [auto-publishing for an npm library](./npm-package
...
+Examples:
+  - AWS account IDs
+  - S3 bucket names
+
+### Secret
+Information that directly causes problems! These should be kept out of VCS completely.
+
+Secrets should be rotated regularly.
+
+If leaked, treat it as a security incident.
```
davidfurey, line 80: Presumably it IS a security incident?

"If leaked, it is a security incident, follow the ..."

akash1810: 👍🏽  updated.

nicl (approved):
All very sensible and helpful changes!

Very very minor (opinionated) nitpick: I'm not so keen on lifting link locations to the bottom. The benefits seem small when links have mostly single use, and it adds a bit of work when refactoring to add/remove links and keep things tidy.

github.md
```diff
@@ -10,23 +18,33 @@ Branch protection should be enabled for the default/release branch of a reposito
...
+- Avoid individual access
+  - Should the individual leave the GitHub organisation, they'll retain access until manually removed
+- Open the repository to the department
+  - [`@guardian/guardian-developers-read`][gh-read] should have read access
+  - [`@guardian/guardian-developers-write`][gh-write] should have write access
+
+### Codeowners
 
-It is recommended to include a `CODEOWNERS` file which references the GitHub team or teams that are responsible for the repository. Where this file is present, it is also recommended to enable the `Require review from Code Owners` option in branch protection.
+Include a [`CODEOWNERS`][gh-codeowners] file which references the GitHub team(s) responsible for the repository.
```
jfsoul, line 32: I'd be tempted to indicate why this is important, because I feel like it is an easy one to ignore for repos where ownership is a bit ambiguous (but that's where it's most needed IMO).

(Feel free to reword, just for example) `On the one hand, this is helpful from a practical point of view, for the team responsible team to receive notifications of contributions and control changes.  On top of that it is also important at the org level to identify owners for every repository so that we can better ensure security and maintenance work is under the remit of a team for all our (production) code.`

github.md
```diff
@@ -10,23 +18,33 @@ Branch protection should be enabled for the default/release branch of a reposito
...
 
-## Codeowners
+### Access
+- Access should be granted to [GitHub teams][gh-teams]
+  - Should an individual leave the GitHub organisation, they'll automatically lose access to any repository in the organisation
+- Avoid individual access
+  - Should the individual leave the GitHub organisation, they'll retain access until manually removed
+- Open the repository to the department
+  - [`@guardian/guardian-developers-read`][gh-read] should have read access
+  - [`@guardian/guardian-developers-write`][gh-write] should have write access
```
jfsoul, line 28: I agree with this, _unless_ one of the later guidelines isn't followed (committing secrets to a private repo). In that scenario, limiting access is warranted, but obviously we should be fixing the actual issue.

Only mentioning because I saw an example recently.

github.md
```diff
@@ -37,3 +55,54 @@ For example, when configuring [auto-publishing for an npm library](./npm-package
 | production | Repositories that are deployed to production                       |
 
 If none of the above topics fit your need, a PR should be opened to add the new topic to this list before use.
+
+## Repository contents
```
jfsoul, line 59: Do you think this sits best in a `security` recommendation or here? Maybe link from the security one?

akash1810:
> Very very minor (opinionated) nitpick: I'm not so keen on lifting link locations to the bottom. The benefits seem small when links have mostly single use, and it adds a bit of work when refactoring to add/remove links and keep things tidy.

Fair point. I find it makes it easier to read a raw markdown file, especially when we don't enforce a max width for files (urls can be massive!).

Saying that, I wouldn't object to the next contributor to this file to revert to inlined links.

github.md
akash1810: Hmm, as these are recommendations, I think this scenario is implicitly covered? (Wait, are implicits good?!)

github.md
akash1810: These [security](https://github.com/guardian/recommendations/blob/main/security.md#sensitive-data) docs? The information covered in both the docs, whilst related, are different. I think cross linking is a good idea.

github.md
akash1810: Great suggestion. I wonder if the doc could be restructured for this reason too. Something like:

```
# Topic
**Recommendation:**
Something short and to the point.

**Reason:**
A paragraph explaining why it's a good thing.
```

As a way of surfacing the recommendation really clearly.

akash1810: ^ this made the doc very busy! I've added an explanation for CODEOWNERS now.

PR #80, opened 2022-02-25
Revamp security recommendations

jfsoul:
## What does this change?
Adds some improvements and additions to the security recommendations. Namely:
- An introductory section, to highlight the importance of security
- More detailed vulnerability management section (but still link to Security HQ for the real detail)
- Rewrite application development section so it introduces the whole topic, rather than just secrets
- Add specific section on secrets, linking to the related github recommendations (with some duplication for ease of discovery)
- Expand AWS section. Adds S3, SGs and patching
- Add patching section
- Other small tweaks

N.B. I don't think this is necessarily the finished article (no such thing), but I think it is a step in the right direction. Notably I think there is a missing reference to SLOs, which we should add relatively soon (probably a separate file)


security.md
```diff
@@ -1,60 +1,172 @@
...
+base images that are end of life.
+
+**Integrate a vulnerability scanning tool and patch vulnerabilities as
+fixes become available.** More guidance can be found on 
+[Security HQ](https://security-hq.gutools.co.uk/documentation/vulnerability-management).
+
 ## HTTPS everywhere
 
-**_All_ external endpoints should be served over HTTPS.**
+**_All_ endpoints should be served over HTTPS.** Use HSTS to enforce it.
```
stiliyana-simeonova, line 169: 👏

security.md
```diff
@@ -1,60 +1,172 @@
...
 
 **Use SSM-Scala for SSH access.**
 [SSM-scala](https://github.com/guardian/ssm-scala#ssm-scala) is a tool
 executing commands on EC2 instances authenticated by IAM credentials.
 The SSM Agent is baked into many of our AMI images and should be
 preferentially used to permanent keys.
 
+### Security Groups
+
+**Security Groups egress/ingress rules should be locked down as much as
```
akash1810, line 124: Should we mention running EC2 instances in the private subnet of the VPC and only the load balancer in the public subnet?

security.md
```diff
@@ -1,60 +1,172 @@
...
+### S3 Buckets
+
+It is surprisingly easy to accidentally make public an S3 bucket or object
+public. There are a number of ways of controlling access, all of which must
+be correctly configured to keep private objects private.
+
+**Prefer using a bucket for a single logical purpose, either public or 
+private.** Mixing concerns makes it more complicated to configure access and
+easier to get it wrong. This allows access to be configured at the bucket 
+level, which is preferred to the object level.
```
akash1810, line 141: Should we recommend granting public access objects should be done via the bucket's policy? (It's a _lot_ simpler than a policy per object). This also reinforces the idea that a bucket is either entirely private or entirely public.

akash1810 (commented):
Looks good! Should we mention provisioning services via CloudFormation is recommended and GuCDK encodes all (I think!) of these best practices by default? That is, use GuCDK to get all/most of this for free.

security.md
jfsoul: I think that is what I'm trying to recommend with `This allows access to be configured at the bucket level, which is preferred to the object level`.

Does that read ok to you - shall I reword it?

security.md
akash1810: Hmm, I don't think we have any "official" docs. We should create them, though out of scope from this PR.

security.md
akash1810: Hmm, not sure why I didn't notice that. Maybe we can split this in two?
  1. Prefer single, logical purpose buckets
  2. Control use of bucket policies to grant access

security.md
jfsoul: 👍 have updated now

security.md
```diff
@@ -1,60 +1,190 @@
...
+
+**Security Groups egress/ingress rules should be locked down as much as
+possible.** There's no reason to allow open access to any EC2 instances; 
+they should ideally be locked down to the load balancer and nothing 
+else. 
+
+**Port 22 access should be removed everywhere.** This is no longer required
+for SSH access via SSM-scala (see above.)
+
+### S3 Buckets
```
jorgeazevedo, line 145: Do you think it's worth mentioning that we should avoid (or be very careful?) when deleting s3 buckets due to risk of a 3rd party taking over the name?

jorgeazevedo (approved):
I think this is a solid update to this document. I actually wasn't aware that this existed, and I think it's a useful reference of what our knowledge of best practice is, which we can update as our understanding evolves.

security.md
jfsoul: 👍  done!

PR #82, opened 2022-03-17
Add VPC section with link to Gateway Endpoints

kenoir:
## What does this change?

Adds a recommendation to use gateway endpoints when talking to S3 & DynamoDB AWS services from a private subnet. 

When using private subnets you can unintentionally incur NAT Gateway costs by talking to AWS services over the internet if these endpoints are not in place and routing tables updated accordingly.


jfsoul (approved):
We probably want to flesh out this section a bit more, but this is good advice as it stands!

PR #83, opened 2022-03-25
Update AWS.md

akash1810:
## What does this change?

Add more detail to the purpose of the `Stack` tag.


davidfurey (approved):

georgeblahblah (approved):

rtyley (approved):
Yup, that's how we use it in the shared AWS Ophan account!

PR #85, opened 2022-04-27
Add more recommendations for VPC's

michaelwmcnamara:
Adds recommendations for VPC's gleaned from https://docs.google.com/document/d/1UAUan8ur3lB1G04x8z3RMWIbOUcGNjEq3WoQGbmDo0I/edit?usp=sharing
to the AWS documentation in the AWS.md file.

Raised as part of https://trello.com/c/C09AXA00/1220-vpc-migration-guide

### DIFF
--- a/AWS.md
+++ b/AWS.md
@@ -38,7 +38,12 @@ EC2
 
 VPC
 ---
-* Ensure you have added the correct [Gateway Endpoints](https://docs.aws.amazon.com/vpc/latest/privatelink/vpce-gateway.html) for the AWS services being accessed from your private subnets to avoid incurring unnecessary networking costs.
+* Ensure you have added the correct [Gateway Endpoints](https://docs.aws.amazon.com/vpc/latest/privatelink/vpce-gateway.html) for the AWS services being accessed from your private subnets to avoid incurring unnecessary networking costs. 
+* Avoid using the default VPC. 
+* Make sure the teams consider whether the account will need to be allocated a block of our IP address space to support peering.
+* If it is likely that AWS resources will need to communicate with our on-prem infrastructure, then contact the networking team to request a CIDR allocation for the VPC.
+* Security of the VPC and security groups must be considered. Limit the inbound traffic to known Guardian or trusted IP ranges. Use SSM-Scala to securely connect to your instances instead of opening SSH ports.
+
 
 ELB
 ---


jfsoul:
Heads up that there's some overlap/duplication with https://github.com/guardian/recommendations/blob/main/security.md#vpc--security-groups

Can we consolidate and cross-link?

michaelwmcnamara:
> 

@jfsoul This is now done. Let me know if it looks ok.

akash1810:
Would it be worth calling out that GuCDK provide's a [construct](https://guardian.github.io/cdk/classes/constructs_vpc.GuVpc.html) for VPCs too?

jfsoul:
+1 to Akash's suggestion, otherwise looks good to me

AWS.md
```diff
@@ -38,7 +38,12 @@ EC2
 
 VPC
 ---
-* Ensure you have added the correct [Gateway Endpoints](https://docs.aws.amazon.com/vpc/latest/privatelink/vpce-gateway.html) for the AWS services being accessed from your private subnets to avoid incurring unnecessary networking costs.
+* Ensure you have added the correct [Gateway Endpoints](https://docs.aws.amazon.com/vpc/latest/privatelink/vpce-gateway.html) for the AWS services being accessed from your private subnets to avoid incurring unnecessary networking costs.
```
jfsoul, line 41: I would consider putting this bullet 4th in this list, above security, as a more natural order of things to consider (`high-level -> ... -> detail`)

nicl (commented):
It might be worth linking to our CDK VPC construct, both because we want to encourage people to use it, and because it has some useful docs of its own (https://github.com/guardian/cdk/blob/main/src/constructs/vpc/vpc.ts#L32-L59).

Beyond that, it would be great if we could explain the reason for things - e.g. *why* is it important not to use the default VPC?

It might also be worth distinguishing between what might be called general-purpose VPCs and ad-hoc ones for specific services. Typically, for general-purposes ones there is one of these per AWS account and they should have the following properties (which our VPC construct ensures/asks for in its docs):

* a public and private subnet per AZ
* A /21 mask (to ensure big enough to meet needs over time)
* a CIDR block provided by Ent Tech to allow for any future peering needs
* gateway endpoints for key services (s3 and dynamo are the starting points here)

Ad-hoc VPCs should be used for specific, very security-sensitive or risky, services - for example our recent Tor/Onion work - but are otherwise discouraged.

michaelwmcnamara:
> It might be worth linking to our CDK VPC construct, both because we want to encourage people to use it, and because it has some useful docs of its own (https://github.com/guardian/cdk/blob/main/src/constructs/vpc/vpc.ts#L32-L59).
> 
> Beyond that, it would be great if we could explain the reason for things - e.g. _why_ is it important not to use the default VPC?
> 
> It might also be worth distinguishing between what might be called general-purpose VPCs and ad-hoc ones for specific services. Typically, for general-purposes ones there is one of these per AWS account and they should have the following properties (which our VPC construct ensures/asks for in its docs):
> 
> * a public and private subnet per AZ
> * A /21 mask (to ensure big enough to meet needs over time)
> * a CIDR block provided by Ent Tech to allow for any future peering needs
> * gateway endpoints for key services (s3 and dynamo are the starting points here)
> 
> Ad-hoc VPCs should be used for specific, very security-sensitive or risky, services - for example our recent Tor/Onion work - but are otherwise discouraged.

Added comments to this effect to the VPC section - also added provided link to the guCDK VPC resources.

AWS.md
```diff
@@ -38,7 +38,20 @@ EC2
 
 VPC
 ---
-* Ensure you have added the correct [Gateway Endpoints](https://docs.aws.amazon.com/vpc/latest/privatelink/vpce-gateway.html) for the AWS services being accessed from your private subnets to avoid incurring unnecessary networking costs.
+
+* Avoid using the default VPC - see [here](https://konekti.us/post/are-you-using-the-aws-default-vpc-stop-now/#:~:text=The%20default%20VPC%20lacks%20the,for%20auditing%20and%20troubleshooting%20purposes.) for a list of reasons why this is a bad idea.
```
nicl, line 42: We can remove the `#...` suffix from this I think. But I also think we should distill the things we care about here and inline it (a bit like Stack Overflow encourages in answers). This helps readers know what we (the Guardian) care about and saves them having to read that entire article - there is a lot of advice in it, some of which we don't (currently) follow. It also ensures our advice remains solid if that website disappears or the URL is taken down/modified.

AWS.md
```diff
@@ -38,7 +38,20 @@ EC2
 
 VPC
 ---
-* Ensure you have added the correct [Gateway Endpoints](https://docs.aws.amazon.com/vpc/latest/privatelink/vpce-gateway.html) for the AWS services being accessed from your private subnets to avoid incurring unnecessary networking costs.
+
+* Avoid using the default VPC - see [here](https://konekti.us/post/are-you-using-the-aws-default-vpc-stop-now/#:~:text=The%20default%20VPC%20lacks%20the,for%20auditing%20and%20troubleshooting%20purposes.) for a list of reasons why this is a bad idea. 
+* Make sure the teams consider whether the account will need to be allocated a block of our IP address space to support peering (see [here](https://docs.aws.amazon.com/vpc/latest/peering/vpc-peering-basics.html) for more info on AWS peering rules.
```
nicl, line 43: Is it worth making this a requirement? Peering is tricky; you may not know you need it up front so better to plan for it anyway. (Also re point below.)

AWS.md
```diff
@@ -38,7 +38,20 @@ EC2
...
+* If it is likely that AWS resources will need to communicate with our on-prem infrastructure, then contact the networking team to request a CIDR allocation for the VPC.
+* If your project uses [guardian/cdk](https://github.com/guardian/cdk) to manage environments and resources, define the VPC using the CDK construct for VPC's [details here](https://github.com/guardian/cdk/blob/main/src/constructs/vpc/vpc.ts#L32-L59). 
+* Ensure you have added the correct [Gateway Endpoints](https://docs.aws.amazon.com/vpc/latest/privatelink/vpce-gateway.html) for the AWS services being accessed from your private subnets to avoid incurring unnecessary networking costs. 
+* For general-purposes, there should be one VPC per AWS account and they should have the following properties (which our VPC construct ensures/asks for in its docs):
```
nicl, line 47: I would be tempted to lead on this in the doc. I.e. use our CDK construct. Doing the right thing and maintaining it over time without this is quite difficult as it requires expertise and also hitting a moving target (as our understanding of best practice evolves over time), whereas CDK provides all of this automatically.

E.g. something like:

---
To follow best practice for VPCs, ensure you have a single CDK-generated VPC in your account that is used to house your applications. You can find the docs for it [here](...).

For exceptional cases, you may want to use the construct to generate further VPCs to isolate specific applications. It is worth discussing with DevX Security and InfoSec if you are thinking of doing this.

You **MUST NOT** use the default VPC. It... (reasons)...

michaelwmcnamara: Good idea!

nicl (approved):
Looks great, thanks @michaelwmcnamara !

PR #86, opened 2022-05-03
Update scala recommendation

mchv:
## What does this change?

The questions came today about what should be the recommended version of the `JDK`.


mxdvl:
(Trying to test a way to surface PRs of general interest)

This PR is a hidden gem 💎 .

https://github.com/guardian/playground/issues/14

PR #87, opened 2022-05-09
Update Java recommendations

nicl:
...to recommend Corretto as the distribution and 11 or newer LTS
versions.

A longer write-up of the rationale can be found here:

https://docs.google.com/document/d/1ZR-YnaXCT5_gLVmTCeGs0mWd3KPaAozPjQK8uUzHZ9w/edit?usp=sharing

### DIFF
--- a/scala.md
+++ b/scala.md
@@ -5,16 +5,16 @@ Scala
  * Ensure you run in production with GC logging enabled
  * Prefer immutability
  * Prefer the latest version of Scala. At the time of writing `3.1.2` is the latest version, so `2.13.8` may be acceptable. if you are still using `2.12.x` you should plan to migrate.
- * If some dependencies are unavailable for the latest version, it is acceptable to use the latest patch version of the previous minor version but try to avoid it. If you have a dependency that isn't available for one of these versions, it's effectively dead and you should plan how to move away from it. 
+ * If some dependencies are unavailable for the latest version, it is acceptable to use the latest patch version of the previous minor version but try to avoid it. If you have a dependency that isn't available for one of these versions, it's effectively dead and you should plan how to move away from it.
  * If you come to a codebase using an older version of Scala, it should be upgraded before any new features are added.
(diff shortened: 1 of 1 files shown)


scala.md
```diff
@@ -5,16 +5,16 @@ Scala
...
  * If you come to a codebase using an older version of Scala, it should be upgraded before any new features are added.
- * Run on a version of the JVM with LTS and with public updates. At the time of writing this means either `JDK17`, `JDK11` or `JDK8`. Recommendation is to use the `JDK11`.
- * Avoid dependency injection in general. 
- 
-[Dependency injection](https://en.wikipedia.org/wiki/Dependency_injection) is a software design pattern that implements [inversion of control](https://en.wikipedia.org/wiki/Inversion_of_control) between a `client` and a `service` for resolving dependencies. The term `injection` refers to a third party (named `the injector`) which is responsible for constructing the services and injecting them into the client. Usual `injectors`are dependency injection frameworks such as `guice`, `dagger`, `macwire`, `spring`. 
+ * Run on a Corretto LTS version of Java, no lower than Java 11. At the time of writing this means Corretto 11 or 17. For tips on upgrading from Java 8, see [here](https://docs.google.com/document/d/1ZR-YnaXCT5_gLVmTCeGs0mWd3KPaAozPjQK8uUzHZ9w/edit?usp=sharing).
```
akash1810, line 10: Is it worth placing this GDoc in GitHub too? Maybe as a wiki if it doesn't naturally fit within this repo?

akash1810 (approved):

scala.md
nicl: I'm not sure the benefit here to be honest. `/recommendations` seems to be for a concise statement of best practice. Google Docs feels better suited to lengthy guides as they are easier to collaborate on and provide easier support for rich document features (TOC, images, etc.).

akash1810: I've been bitten in the past where GitHub referred to a GDoc that no longer existed because the owner had left and didn't change ownership of it. That's harder to do with VCS.

Not a blocker.

nicl: Ah yeah, that is horrible! The doc is in the shared drive at least, so hopefully should be safe for a while.

PR #88, opened 2022-05-13
Add recommendation for logging in .json format to scala recommendations

michaelwmcnamara:
## What does this change?
When logging in .json format for a scala app, it is necessary to use net.logstash.logback.encoder.LogstashEncoder.
When using this with Play 8.x you are exposed to a Denial of Service vulnerability in the Jackson library used by this version of Play.
This change documents this problem and adds details for the current workaround to resolve this issue.

Trello card: https://trello.com/c/tiXyFoIv/1259-support-ingesting-json-logs-and-define-best-practice-on-this

### DIFF
--- a/npm-packages.md
+++ b/npm-packages.md
@@ -93,11 +93,17 @@ The CommonJS version should be referenced by the `main` field, and the TypeScrip
 
 #### Continuous delivery
 
-Use [Semantic Release](https://github.com/semantic-release/semantic-release) in a GitHub action.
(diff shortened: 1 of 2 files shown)


akash1810:
I'm not sure we're at the recommendation stage yet, and should maybe add this to the proposal doc instead for now? cc @guardian/devx-operations .

johnduffell (commented):
probably not needed any more now that play 2.8 is out of service and we should all be on play 3.0

PR #89, opened 2022-05-25
Add note about infosec access

davidfurey:
## What does this change?

Adds a note about infosec AWS access if Janus is unavailable.


stiliyana-simeonova (approved):
👍

davidfurey:
Thanks

PR #90, opened 2022-06-09
Add section to config on injecting on app start

jamesgorrie:
## What does this change?

Adds a section to the config about ensuring that we make the config editable on deploy, rather than build or any other process.

Comes from a conversation with @OllysCoding @davidfurey and others, and thought it was worth capturing.

## How to test

Read the update to the document, and see if it makes sense to future engineers.

Future engineers understand this.


davidfurey (approved):
Thanks for adding this

PR #91, opened 2022-06-20
Add broad expectations for production application creation and maintenance

jfsoul:
## What does this change?
This adds a slightly multi-purpose recommendations document around production service creation and maintenance. There are many recommendations in this repository already which clearly apply to production services, but nowhere do we spell out the minimum expectations before a prototype can be promoted/deployed to production. The lack of that guidance leads to some projects that currently do not meet those expectations.

We also require teams to perform regular maintenance tasks so that these expectations, and others, continue to be met. I've added those here too, as they seem related - plus how we expect projects with no official engineering team owner, to be treated.

Engineers should be clear on the dos and don'ts of production services. Not all these expectations are measured at the moment, but some are, and we will continue to monitor those.

## Other notes/questions
- The lists in the new document are probably not complete yet, ideas welcome!
- Should this be split into multiple documents?
- Is the overlap with other recommendations acceptable?


ownership.md
```diff
@@ -0,0 +1,54 @@
...
+- CI/CD should be employed
+    - An appropriate testing strategy should be considered. We do not aim for a specific % of test coverage, but important business logic should be unit tested
+    - For CI, we use GitHub Actions and [Teamcity](https://teamcity.gutools.co.uk/)
+    - For most (non-library) projects, deployment will be done via [Riff-Raff](https://riffraff.gutools.co.uk/)
+
+### Security
+- A basic [security](./security.md) assessment should be performed to understand the risks and available controls. E.g. 
+authentication, network security, encryption, secret management. Expert guidance from outside the team should sought for high risk applications (e.g. processing user data)
+- Any dependency manifest files should be scanned using [Snyk Open Source](https://security-hq.gutools.co.uk/documentation/snyk)
+- Internal tools should be behind Google Auth. (Network-layer restrictions may also be recommended based on the context)
```
kelvin-chappell, line 22: Perhaps a link to an example where this is in use or documentation.

ownership.md
```diff
@@ -0,0 +1,54 @@
...
+
+### Infrastructure
+- AWS infrastructure should *not* be deployed to the Developer Playground account
+- All new AWS infrastructure should be defined using [CDK](https://github.com/guardian/cdk), and legacy services should be migrated when possible
+- Infrastructure costs should be estimated and within existing budget (unless otherwise approved by Head/Director of Engineering)
+
+### Observability and support
+- Monitoring and alerting should exist to ensure the owner is notified when the core functionality of a service is unavailable or impaired
+- Application logs and (if necessary/applicable) telemetry data should be [shipped](https://github.com/guardian/deploy-tools-platform/tree/main/elk), so that service impairment can be debugged effectively for common problems (e.g. application errors, resource constraints)
+- For high priority applications, a runbook should be created to describe to other engineers how to debug and address service impairment incidents
```
kelvin-chappell, line 33: Perhaps a link to a good example of a runbook or guidelines on writing runbooks

ownership.md
```diff
@@ -0,0 +1,54 @@
...
+
+## High level maintenance expectations
+
+The broad expectations of ongoing maintenance of a production application are:
+1. Security vulnerabilities are addressed *as a priority*
+1. Dependencies are kept up to date, so that security patching requirements can be met efficiently
```
kelvin-chappell, line 39: Is it worth mentioning Dependabot and the internal Scala Steward lists here?

ownership.md
```diff
@@ -0,0 +1,54 @@
...
+1. Unused functionality is removed
+1. Impaired functionality is addressed, with priority determined by its importance
+1. Costs are monitored. [Baseline monitoring](https://github.com/guardian/aws-cost-management) is added globally
+1. Architecture/design is reviewed periodically to ensure adherence to best practices and any SLA/SLOs at the team or org level
+
+### Orphaned Projects
+In the vast majority of cases we address these by assigning ownership of source code to an official engineering team. Official engineering teams have useful properties for this task, like resourcing oversight and planning ceremonies.
```
kelvin-chappell, line 46: Is the recommendation to adopt codeowners here or is that too prescriptive?

ownership.md
jfsoul: We have a plan to overhaul the documentation on dependency management separate to this. I'll add a note to the card so that we come back here and add a link to it.

jfsoul:
@kelvin-chappell thank you for the review! Can you think of any areas we might be missing?

github.md
```diff
@@ -26,11 +26,15 @@ Generally, repositories should be open to the department via these teams:
...
 - [`@guardian/guardian-developers-write`][gh-write] should have write access
 
-### CODEOWNERS
-Include a [`CODEOWNERS`][gh-codeowners] file which references the [GitHub team(s)][gh-teams] responsible for the repository.
+### Collaborators and CODEOWNERS
+Include at least one GitHub team as a collaborator with admin access. *At an organisational level, this identifies owners for every repository, so that we can better ensure security and maintenance work is under the remit of a team for all our (production) code.*
+
+In general, avoid giving individual access to repositories and prefer teams.
+
+Include a [`CODEOWNERS`][gh-codeowners] file which references the [GitHub team(s)][gh-teams] responsible for different parts of the repository. *This allows the team responsible team to receive notifications of contributions and review changes.*
```
akash1810, line 34:
```suggestion
Include a [`CODEOWNERS`][gh-codeowners] file which references the [GitHub team(s)][gh-teams] responsible for different parts of the repository. *This allows the responsible team to receive notifications of contributions and review changes.*
```

ownership.md
```diff
@@ -0,0 +1,56 @@
...
+- A basic [security](./security.md) assessment should be performed to understand the risks and available controls. E.g. 
+authentication, network security, encryption, secret management. Expert guidance from outside the team should sought for high risk applications (e.g. processing user data)
+- Any dependency manifest files should be scanned using [Snyk Open Source](https://security-hq.gutools.co.uk/documentation/snyk)
+- Internal tools should be behind Google Authentication 
+    - A helper exists for [Scala](https://github.com/guardian/play-googleauth) and authentication can be added to an [ALB directly](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/listener-authenticate-users.html)
+    - Network-layer restrictions may also be recommended based on the context
+- All new 3rd party software / vendors used as part of a service should have approval via the [official process](https://spike.gnmremote.com/content/178/requesting-software-for-your-gnm)
+
+### Infrastructure
+- AWS infrastructure should *not* be deployed to the Developer Playground account
```
akash1810, line 28: Is it worth saying why too? Is it because the playground account can be cleared at anytime?

ownership.md
```diff
@@ -0,0 +1,56 @@
...
+1. Unused functionality is removed
+1. Impaired functionality is addressed, with priority determined by its importance
+1. Costs are monitored. [Baseline monitoring](https://github.com/guardian/aws-cost-management) is added globally
+1. Architecture/design is reviewed periodically to ensure adherence to best practices and any SLA/SLOs at the team or org level
+
+### Orphaned Projects
+In the vast majority of cases we address these by assigning [ownership](./github.md#collaborators-and-codeowners) of source code to an official engineering team. Official engineering teams have useful properties for this task, like resourcing oversight and planning ceremonies.
+
+However there are some repositories (containing production software) we have created that have no obvious official engineering team owner. Instead we expect a small group of volunteers to be responsible for the above maintenance tasks. We expect these volunteers to:
+- Create a GitHub team and add volunteers as team members. Notify the Developer Experience team of this (the team should be admins of all relevant GitHub repos)
```
akash1810, line 51: We suggest an EM be involved in this process. As EMs have GH admin rights, can we drop the need to notify DevX here?

akash1810 (commented):
LGTM!

ownership.md
jfsoul: Unfortunately we still have a need to be notified of new github teams that can be considered maintainers of a repository. It allows us to exclude teams like `all` from our monitoring and ensure somebody is responsible.

ownership.md
jfsoul: Plus security concerns. I think we essentially treat the account as a test environment, so adding production services/data would be a concern. Access to the account is also very wide.

PR #93, opened 2022-07-15
Encourage devs to check their PRs after merging

rtyley:
As a dev it's very easy, as soon as you merge a PR, to think of it as completed, and move on to the next thing. However, this is actually the most important time to check that the change is *working*!

This recommendation update aims to encourage devs to extend their mental model of the lifecycle of a PR a little further, post-merge, to take a little time to check on the change.


davidfurey (approved):
🥳

PR #96, opened 2022-09-02
Add page with recommendations for AWS Embedded Metric format

michaelwmcnamara:
## What does this change?
This adds a recommendation page detailing how to use AWS Embedded Metrics format.
This is the recommended way to write to Metrics from Lambda functions or Container-based apps as it has many advantages over the other methods [Full details on that here](https://docs.google.com/document/d/1cL_t5NhO8J9Bwiu4rghoGh8i_um_sXDyKuq4COhdLEc/edit?usp=sharing)
It is however a technique that is not well known.
Recommendations cover what prerequisites you need to use this method as well as details on the format used to embed a metric within a logging request. It also contains a code snippet showing a sample logging request.

### DIFF
--- a/AWS-embedded-metrics.md
+++ b/AWS-embedded-metrics.md
@@ -1,0 +1,52 @@
+AWS Embedded Metrics
+--------------------
+* AWS Embedded metrics are method for capturing Cloudwatch metrics as part of a logging request.
+* This is an ideal solution for generating metrics for Lambda functions that will track historical data.
+* This does not work at all for our EC2 apps as their logs do not pass through Cloudwatch.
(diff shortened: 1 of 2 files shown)


nicl (approved):
Looks great. I wonder if it is worth calling the file `monitoring` or `metrics` or something more general? Although this is really a broader issue with recommendations than this PR - the lack of structure/hierarchy.

jacobwinch:
Great work @michaelwmcnamara 👍 

> I wonder if it is worth calling the file monitoring or metrics or something more general?

I agree that a more general name might be better here - perhaps `lambda-monitoring`, since this technique/recommendation really only applies to Lambdas in our context?

michaelwmcnamara:
Great recommendation @jacobwinch @nicl - I'll update then merge :-)

PR #97, opened 2022-09-09
Node.js recommendations for client side projects

SiAdcock:
## What does this change?

Specify recommendations for Node.js projects

To me, it always makes sense to use the latest even-numbered version of Node, unless there is a good reason not to. 

Odd-numbered versions become unsupported more quickly than even-numbered.

Although there are newer version managers than nvm, the `.nvmrc` file is still the most prevalent and well-supported file for indicating Node.js version. `fnm`, `asdf` and `nvm` all support `.nvmrc` files

### DIFF
--- a/client-side.md
+++ b/client-side.md
@@ -37,6 +37,7 @@ See the separate [npm-packages.md](./npm-packages.md).
 
 -   Prefer using the _Current_ even-numbered [version of Node.js](https://nodejs.org/en/about/releases/)
 -   Use an `.nvmrc` file to indicate the exact version of Node.js your project uses
+    - *Note:* `asdf` users must configure their tool to use [legacy version files](https://github.com/asdf-vm/asdf-nodejs#nvmrc-and-node-version-support)
 
 
 ## Strategies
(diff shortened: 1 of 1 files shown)


client-side.md
```diff
@@ -33,6 +33,13 @@ See the separate [npm-packages.md](./npm-packages.md).
     otherwise.
 -   Prefer locking your dependencies (direct and transitive) using a lockfile rather than committing them into Git.
 
+## Node.js
+
+-   Prefer using the _Current_ even-numbered [version of Node.js](https://nodejs.org/en/about/releases/)
```
davidfurey, line 38: It would be helpful to include your PR comment "Odd-numbered versions become unsupported more quickly than even-numbered." here to explain why you encourage even-numbered releases

client-side.md
```diff
@@ -33,6 +33,14 @@ See the separate [npm-packages.md](./npm-packages.md).
...
+    - Odd-numbered versions become unsupported more quickly than even-numbered
+-   Use an `.nvmrc` file to indicate the exact version of Node.js your project uses
```
sndrs, line 40:
```suggestion
-   Use an `.nvmrc` file to indicate the exact version of Node.js your project uses
	- It's the only node version file that is supported by all current node version management tools
```

PR #98, opened 2022-10-13
add a section on NPM `peerDependencies`

sndrs:
## What does this change?

adds a section on `peerDependencies` in npm packages, in particular around how to handle them in monorepos

### DIFF
--- a/npm-packages.md
+++ b/npm-packages.md
@@ -104,6 +104,138 @@ The CommonJS version should be referenced by the `main` field.
 }
 ```
 
+### `peerDependencies`
+
+#### `peerDependencies` ranges should be a wide as possible
+
+This ensures compatibility with the maximum number of installations.
+
+#### Pin the lowest possible version of `peerDependencies` in `devDependencies`
+
+This ensures your library is developed against the sparest possible version of its dependencies.
+
+##### Example
+
+```js
+// my-lib/package.json
+{
+	"name": "my-lib",
+	"version": "1.0.0",
+    "devDependencies": {
+        "your-lib": "1.2.3"
+    },
+    "peerDependencies": {
+        "your-lib": "^1.2.3"
+    }
+}
+```
+
+#### Changes to `peerDependencies` ranges are breaking
+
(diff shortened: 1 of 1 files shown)


sndrs:
cc @padolsey

padolsey:
Thanks so much for putting this together! Makes sense – I think – tho could it use better example lib names? My brain plays tricks on me with "their-lib", "my-lib", "lib-a" etc. (A bit like when people use double/triple negatives in sentences 😅). Perhaps something typical like "App" and a dependency, "Banner", or something more Guardian-esque, would be more intuitive?

Also, recommending that consumers figure out the "earliest version" of a dependency they can use seems optimistic. How can we incentivize them to do that? Can tooling/testing help? People tend to just stick with "north of" (^) whatever version they first installed. Maybe that's sufficient tho? (Disclaimer: still trying to get my head around this stuff)

sndrs:
> could it use better example lib names? 

agreed

sndrs:
> recommending that consumers figure out the "earliest version" of a dependency they can use seems optimistic

is this referring to this bit?

> Pin the lowest possible version of `peerDependencies` in `devDependencies`

this is intended for the authors of packages that have peerdeps, to ensure they build and test their package locally with the earliest version of the package they say they can use (i.e. if you specify a peerdep of `"moment": "^2.1.0"`, install `moment@2.1.0` as a dev dep)

is this the part you meant? if so that def needs rewording – it's not intended for consumers!

padolsey:
Ah I see now - I think it is indeed clear enough. And thank you for popping in the more realistic names; it definitely helps digesting this all. 💯

I'll approve : ) (thanks for patience!) but want to drop a micro-concern: Dependency juggling is complex and I reckon people could still potentially get confused. If everyone, @ob6160, @tjsilver, @joecowton1, can take a quick look at this recommendation, just as a final litmus test. We need to make sure it's intuitive for any potential projects coming into the monorepo.

padolsey (approved):

tjsilver (approved):
This reads well and makes sense to me 👍

PR #100, opened 2023-01-11
Prefer systemd timers to crontab for scheduled jobs

rtyley:
Suggesting a new recommendation here - while debugging the issues behind https://github.com/guardian/grid-ftp/pull/13, @sihil pointed out that systemd timers have advantages over crontab jobs! They do require a bit more setup (eg separate `.service` & `.timer` files), but it sounds like we should probably prefer them, and let people know about them?

At the moment, GitHub search seems to show only a few places where we use [systemd-timers](https://github.com/search?q=org%3Aguardian+systemd+timer&type=code), whereas there are _lots_ of [`crontab`](https://github.com/search?p=2&q=org%3Aguardian+crontab&type=Code).

### DIFF
--- a/scheduled-jobs.md
+++ b/scheduled-jobs.md
@@ -1,0 +1,23 @@
+# Scheduled jobs - tasks that run periodically
+
+## Cron expressions
+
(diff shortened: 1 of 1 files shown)


kelvin-chappell:
Not sure if it's in the scope of a recommendation but do you think a link to an example where this has been configured would help take-up?  Or perhaps a simple set of steps?

rtyley:
> do you think a link to an example where this has been configured would help take-up?

Thanks @kelvin-chappell, I've created https://github.com/guardian/editorial-ftp/pull/22 to see what switching from crontab to a systemd timer would look like, have not fully deployed it yet to check it works!

sihil:
Another example in the form of the elk-housekeeping files here: https://github.com/guardian/deploy-tools-platform/tree/main/elk/config

davidfurey (approved):

PR #103, opened 2023-02-13
Initial draft of recommendations on CDNs

davidfurey:
## What does this change?

Adds some recommendations on using CDNs. I don't think this is documented elsewhere.  I suspect there is a lot more useful info we could include but this seems like a reasonable starting point.

Please do question these recommendations if you disagree if them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ provides simpler scalability and greater robustness
 -   [Client Side](client-side.md)
 -   [Coding with empathy](coding-with-empathy.md)
 -   [Config](config.md)
+-   [Content Delivery Network (CDN)](cdn.md)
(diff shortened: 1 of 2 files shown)


akash1810 (commented):
Should we also recommend [Fastly shielding](https://docs.fastly.com/en/guides/shielding)?

philmcmahon (commented):
Add sections on shielding/logging

philmcmahon:
Looks great to me. Following @akash1810 's suggestion I've added a shielding bit. Have also put a section on logging in - might be overkill but it comes up fairly frequently. 

Feel free to suggest/make any changes

davidfurey:
Thanks Phil.  I've added a note about IAM for logging

akash1810 (approved):

PR #105, opened 2023-03-01
Add notes on Scala Steward

kelvin-chappell:
## What is being recommended?
The recommendation is that Scala Steward is used for management of Scala dependencies.  I've called this *Continuous* dependency management but it isn't quite that.  That's the general direction of travel though.

## What's the context?
I've been using Scala Steward for a while now, to the extent that I'd feel a bit lost without it.  
It's in use on a wide range of our repos and we have an in-house system for managing it.  So it seems stable and conventional enough to warrant a recommendation.  We aren't, as far as I'm aware, explicitly supporting any competitor systems.
This just brings together in one place the rationale for using it and some advice around it.

---

@rtyley as you and @jfsoul have done all the work to support Scala Steward in Guardian repos I wonder if there's anything else you think should be mentioned in this recommendation?  I felt that the git-blame issues might fit better in a separate `scalafmt` recommendation.  And I haven't said too much about whether grouping dependencies is good or bad.  Perhaps calling that section *bundling* dependency updates might be misleading.  I'm not sure.

### DIFF
--- a/scala.md
+++ b/scala.md
@@ -1,5 +1,4 @@
-Scala
------
+# Scala
 
  * Always specify an explicit sbt.version
  * Ensure you run in production with GC logging enabled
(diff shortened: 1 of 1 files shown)


jfsoul:
👋 I'll refrain from commenting too much as my opinion no longer counts! However, I will say:
- Around the time I left, grouping updates became available natively in Scala Steward, which I think makes using the custom PR batching workflow I authored difficult to justify.
- A few other recommendations cross over with this added section. Think about updating, linking from or consolidating in some way with the ownership, dependencies and security recommendations.
All the best 😊

kelvin-chappell:
Thanks for your comments, @jfsoul!

> * Around the time I left, grouping updates became available natively in Scala Steward, which I think makes using the custom PR batching workflow I authored difficult to justify.

I wasn't aware of this.  I'll have a look.

> * A few other recommendations cross over with this added section. Think about updating, linking from or consolidating in some way with the ownership, dependencies and security recommendations.

This is a good point.  I'll have a look around at what already exists.

kelvin-chappell:
Now slightly updated and I'll add a link to https://github.com/guardian/security-hq/blob/main/hq/markdown/vulnerability-management.md#vulnerabilities-via-3rd-party-libraries when this has been merged.

NovemberTang (approved):

PR #59, opened 2021-05-11
Create dependencies.md

mxdvl:
## How to test

Already present in:
- `consent-management-platform` since [c3c18492c](https://github.com/guardian/consent-management-platform/commit/c3c18492c541a5f95962e76c88479999bd5012a6)

- `commercial-core` since [26815b77](https://github.com/guardian/commercial-core/pull/285/commits/26815b772a461608667839d126c46abdbbc564e7)

## How can we measure success?

Allow shared dependencies to get security fixes, and let the responsibility lie with the consumer applications.

## Have we considered potential risks?

Added known issues/packages which are not respecting semantic versioning.

## Accessibility

- [ ] [Tested with screen reader](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/558398)
- [ ] [Navigable with keyboard](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/39894d)
- [ ] [Colour contrast passed](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/92b913)
- [ ] [The change doesn't use only colour to convey meaning](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/29032f)

PR #114, opened 2023-05-02
Clarify JS, NPM & Dependencies recommendations

mxdvl:
## What is being recommended?

The `dependencies`, `npm-packages` and `client-side` recommendations have been reworked so they no longer contradict each other.

## What's the context?

Follow-up on #53 & #59.

### DIFF
--- a/client-side.md
+++ b/client-side.md
@@ -22,16 +22,17 @@ See the separate [npm-packages.md](./npm-packages.md).
 
 -   Have an [.editorconfig](http://editorconfig.org/) file in your repo.
 -   Automate linting as part of your test phase (e.g. jshint, eslint,
-    csslint).
-    -   Use the [departmental linting rules](https://github.com/guardian/configs)
+    csslint) using the department linting rules:
+    - [@guardian/eslint-config](https://github.com/guardian/csnx/tree/main/libs/%40guardian/eslint-config)
+    - [@guardian/eslint-config-typescript](https://github.com/guardian/csnx/tree/main/libs/%40guardian/eslint-config-typescript)
+    - [@guardian/tsconfig](https://github.com/guardian/csnx/tree/main/libs/%40guardian/tsconfig)
+    - [@guardian/prettier](https://github.com/guardian/csnx/tree/main/libs/%40guardian/prettier)
 -   Don’t enforce coding styles in pull requests. Add linting dot files
     (e.g. .jshintrc) to your project instead.
 
 ## Dependencies
 
--   If starting a new project, favour ES6 modules unless you can justify
-    otherwise.
--   Prefer locking your dependencies (direct and transitive) using a lockfile rather than committing them into Git.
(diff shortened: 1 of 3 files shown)


npm-packages.md
```diff
@@ -39,35 +39,34 @@ import ts from "rollup-plugin-ts";
 import pkg from "./package.json";
 
 export default [
-	{
```
sndrs, line 42: spaces instead of tabs goes against [the recommendations of the guardian prettier config](https://github.com/guardian/csnx/blob/12fdc644ac963611b8214d5b349f9184a26e6839/libs/%40guardian/prettier/index.js#L12-L17) :)

(we should add a prettier config to this repo to catch that)

mxdvl: Haha wow, I got this formatting by actually running prettier. Will revert!

tjmw: TIL that prettier supports formatting code in markdown (and has done for years!) https://prettier.io/blog/2017/11/07/1.8.0.html

sndrs: yeah does a lot by default https://github.com/guardian/csnx/blob/main/.lintstagedrc.js#L2

mxdvl: Draft: https://github.com/guardian/recommendations/pull/117

client-side.md
```diff
@@ -22,16 +22,17 @@ See the separate [npm-packages.md](./npm-packages.md).
...
+    - [@guardian/tsconfig](https://github.com/guardian/csnx/tree/main/libs/%40guardian/tsconfig)
+    - [@guardian/prettier](https://github.com/guardian/csnx/tree/main/libs/%40guardian/prettier)
 -   Don’t enforce coding styles in pull requests. Add linting dot files
     (e.g. .jshintrc) to your project instead.
 
 ## Dependencies
 
--   If starting a new project, favour ES6 modules unless you can justify
-    otherwise.
--   Prefer locking your dependencies (direct and transitive) using a lockfile rather than committing them into Git.
```
tjmw, line 34: I think we lost these two recommendations along the way? Are they so obvious now we don't need to spell them out, or would it be useful to capture them still?

tjmw (approved):
Nice! 👍

client-side.md
mxdvl: 1. I both considered that at this point[ES6 (2015) is the standard](https://caniuse.com/es6), but also did not really fit below the **Dependencies** header. I would suggest a new **Modules** header if we want to capture more updated recommendations.
2. `pnpm`, `yarn` and `npm` now all create lockfiles by default. [We already mention lockfiles in `dependencies.md`](https://github.com/guardian/recommendations/blob/6dbdefa26502a72621279c53df51df44b28ecb99/dependencies.md?plain=1#L35)

Do you think this needs to be addressed in this PR?

tjmw: 👍 Good points. No need to change anything in this PR.

PR #115, opened 2023-05-02
Pin JavaScript dependencies

mxdvl:
## What is being recommended?

Recommend pinning dependencies in JavaScript applications. This ensures consistency and reliability.

## What's the context?

https://docs.renovatebot.com/dependency-pinning/#why-pin-dependencies

PR #120, opened 2023-05-10
fix: add reason and examples for pinning deps to patch

jamesgorrie:
Extension from https://github.com/guardian/recommendations/pull/115
- Adds reason and link for pinning deps (via @mxdvl)
- Adds an example for more context


mxdvl (approved):
I like it!

Note that `Good` and `bad` have different capitalisation

PR #124, opened 2023-05-31
feat: Generate the best practice Markdown

akash1810:
This follows on from #119.

## What is being ~recommended~ changed?
Use [Deno](https://deno.com/) for a lightweight script to generate a Markdown table, and inject it into best-practice.md.

The TypeScript type-system allows for a schema to be created, allowing each best practice to render the same way.

## What's the context?
In #119 I commented:

> Writing markdown tables by hand is not fun!

This is still true! I also commented:

> I'll think of a lightweight way to introduce auto-generation of these tables. Initial thinking is YAML -> Markdown. Use of YAML allows for comments, and a schema for "type-checking".

Alas, converting YAML to a Markdown table wasn't quite as simple as this Deno script. The script still provides the same benefits of allowing for comments, and type-checking.

There are two downsides to this solution:
1. A requirement to install Deno. However, I'd suggest this is easier to maintain vs. the bash needed for the YAML version.
2. An understanding of TypeScript. However, I'd suggest this is minimal as only `definitions.ts` needs updating, and the IDE should pick up the added doc strings to offer IntelliSense.

Lastly, in #119 @mxdvl [commented](https://github.com/guardian/recommendations/pull/119#issuecomment-1539742128) that CSVs are supported by GitHub, and better-suited at tabular data. Whilst this is true, Markdown tables offer richer formatting, such as links.

### DIFF
--- a/best-practice/definitions.ts
+++ b/best-practice/definitions.ts
@@ -1,6 +1,6 @@
 import { IAllBestPractice, IBestPractice } from "./types.ts";
 
-const github: IBestPractice[] = [
+const github: readonly IBestPractice[] = [
   {
     name: "Default Branch Name",
     owner: "[@guardian](https://github.com/orgs/guardian/teams/all)",
@@ -9,9 +9,9 @@ const github: IBestPractice[] = [
     howToCheck: "Manual. View the repository on https://github.com",
     howToExempt: "N/A",
   },
-];
+] as const satisfies readonly IBestPractice[];
 
-const aws: IBestPractice[] = [
+const aws: readonly IBestPractice[] = [
   {
     name: "Resource Tagging",
     owner:
@@ -21,7 +21,7 @@ const aws: IBestPractice[] = [
     howToCheck: "TBD",
     howToExempt: "N/A",
   },
-];
+] as const satisfies readonly IBestPractice[];
 
 export const AllBestPractices: IAllBestPractice = {
   GitHub: github,
--- a/best-practice/definitions.ts
+++ b/best-practice/definitions.ts
@@ -1,0 +1,29 @@
+import { IAllBestPractice, IBestPractice } from "./types.ts";
+
+const github: IBestPractice[] = [
+  {
+    name: "Default Branch Name",
+    owner: "[@guardian](https://github.com/orgs/guardian/teams/all)",
+    description:
+      "The default branch name should be `main`.<br>See the [master-to-main tool](https://github.com/guardian/master-to-main/blob/main/migrating.md).",
+    howToCheck: "Manual. View the repository on https://github.com",
+    howToExempt: "N/A",
+  },
+];
+
+const aws: IBestPractice[] = [
+  {
+    name: "Resource Tagging",
+    owner:
+      "[DevX Operations](https://github.com/orgs/guardian/teams/devx-operations)",
+    description:
+      "AWS resources should be tagged (where supported) with `Stack`, `Stage`, and `App`.<br>This aids service discovery, and cost allocation.",
+    howToCheck: "TBD",
+    howToExempt: "N/A",
+  },
+];
+
+export const AllBestPractices: IAllBestPractice = {
+  GitHub: github,
+  AWS: aws,
+};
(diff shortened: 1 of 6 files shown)


best-practice/index.ts
```diff
@@ -0,0 +1,67 @@
...
+
+import * as path from "https://deno.land/std/path/mod.ts";
+import { markdownTable } from "npm:markdown-table";
+import { AllBestPractices } from "./definitions.ts";
+
+const markdownFile = path.join(
+  path.dirname(path.fromFileUrl(import.meta.url)),
+  "..",
+  "best-practices.md",
+);
```
mxdvl, line 15: You can also do `new URL(import.meta.resolve('../best-practices.md')`

best-practice/definitions.ts
```diff
@@ -0,0 +1,29 @@
...
+const github: IBestPractice[] = [
+  {
+    name: "Default Branch Name",
+    owner: "[@guardian](https://github.com/orgs/guardian/teams/all)",
+    description:
+      "The default branch name should be `main`.<br>See the [master-to-main tool](https://github.com/guardian/master-to-main/blob/main/migrating.md).",
+    howToCheck: "Manual. View the repository on https://github.com",
+    howToExempt: "N/A",
+  },
+];
```
mxdvl, line 12: I would consider using a constant assertion to narrow this object's type to an exact version:

```ts

const github = [
/* the object */
] as const satisfies IBestPractice;
```

best-practice/index.ts
```diff
@@ -0,0 +1,67 @@
...
+
+const file = await Deno.readTextFile(markdownFile);
+
+const startMark = "<!-- csvstart -->";
+const endMark = "<!-- csvend -->";
```
mxdvl, line 20: Might be a little confusing if we're not using CSV…

mxdvl (commented):
How nice 🦕

best-practice/index.ts
akash1810: 🤦🏽  will update.

best-practice/definitions.ts
akash1810: `as const satisfies IBestPractice[];` fails to compile as the `as const` makes everything readonly:

```log
error: TS1360 [ERROR]: Type 'readonly [{ readonly name: "Default Branch Name"; readonly owner: "[@guardian](https://github.com/orgs/guardian/teams/all)"; readonly description: "The default branch name should be `main`.<br>See the [master-to-main tool](https://github.com/guardian/master-to-main/blob/main/migrating.md)."; readonly howToCheck: "Ma...' does not satisfy the expected type 'IBestPractice[]'.
  The type 'readonly [{ readonly name: "Default Branch Name"; readonly owner: "[@guardian](https://github.com/orgs/guardian/teams/all)"; readonly description: "The default branch name should be `main`.<br>See the [master-to-main tool](https://github.com/guardian/master-to-main/blob/main/migrating.md)."; readonly howToCheck: "Ma...' is 'readonly' and cannot be assigned to the mutable type 'IBestPractice[]'.
] as const satisfies IBestPractice[];
```

I'm not sure I understand the advantage of using `satisfies` here. Could you elaborate please?

best-practice/index.ts
```diff
@@ -0,0 +1,67 @@
...
+
+// Find text between markers and replace it with the start marker
+const re = new RegExp(`${startMark}(.|\n)*${endMark}`, "m");
+const resetFile = file.replace(re, startMark);
+
+// replace the now singular start marker with the start marker, content, and end marker
+const updatedFile = resetFile.replace(
+  startMark,
+  [startMark, ...markdownContent, endMark].join("\n"),
+);
```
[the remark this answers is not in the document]
akash1810: No real reason. Updated to one step now.

best-practice/definitions.ts
mxdvl: Using `satisfies` ensures that the input data is both narrowed but also adheres to the expected shape. Without it, a future contributor might add an unexpected key in on of the arrays' objects, and `as const` would not complain.

For data that is truly immutable I think it's the best of both declaring interfaces and getting an exact type.

In this case we would probably need to make the interface `readonly`, which is more correct anyways as we do not want any method to inadvertently mutate it.

mxdvl (approved):

PR #125, opened 2023-06-05
Add recommendation around RFCs

sndrs:
## What is being recommended?

add some best practice and rationale around RFCs.

## What's the context?

i've encountered some confusion about what RFCs are for, what they entail and what's expected of commenters.

this is an attempt to settle on a common understanding

### DIFF
--- a/RFCs.md
+++ b/RFCs.md
@@ -21,6 +21,6 @@ This will help readers to more easily understand your rationale.
 Comments on Google docs or GitHub discussions are good ways of collecting people's thoughts alongside your original proposal.
 
 ## How to comment on an RFC
-A RFC is a request for help – someone is asking for your insight and the benefit of your experience. Be candid but be kind. Ultimately, the aim is to collaborate until collectively, we arrive at the best possible version of the idea.
+An RFC is a request for help – someone is asking for your insight and the benefit of your experience. Be candid but be kind. Ultimately, the aim is to collaborate until collectively, we arrive at the best possible version of the idea.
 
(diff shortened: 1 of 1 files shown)


akash1810 (commented):
Should we also describe _where_ to write the RFC? Google Docs? A file in the repo? An issue? A discussion?

IMO it should be somewhere that won't get lost, and is accessible by all that it needs to be.

RFCs.md
```diff
@@ -0,0 +1,26 @@
...
+1. The problem you want to solve
+2. Why the status quo does not address the problem
+3. A proposed solution
+
+This will help readers to more easily understand your rationale.
+
+Comments on Google docs or GitHub discussions are good ways of collecting people's thoughts alongside your original proposal.
+
+## How to comment on an RFC
+A RFC is a request for help – someone is asking for your insight and the benefit of your experience. Be candid but be kind. Ultimately, the aim is to collaborate until collectively, we arrive at the best possible version of the idea.
```
akash1810, line 24: We've a mix of grammar in this document, with use of "an RFC" and "a RFC"...

akash1810 (commented):
Worth linking to some previous examples?

sndrs:
> Should we also describe _where_ to write the RFC? Google Docs? A file in the repo? An issue? A discussion?
> 
> IMO it should be somewhere that won't get lost, and is accessible by all that it needs to be.

i mentioned using docs or github discussions, but only really in passing. it kind of feels like some are better in docs (eg process-based proposal where non-engineers may want to contribute) and others in github (pure engineering decisions where you want to connect PRs, commits, discussions etc).

would that be a good way of framing it do you think?

kenoir (approved):
Neato 👌

PR #128, opened 2023-06-23
Add some GitHub best practices

akash1810:
## What is being ~recommended~ changed?
Tabulating the information on https://github.com/guardian/recommendations/blob/main/github.md for easy referencing.

See the final result [here](https://github.com/guardian/recommendations/blob/aa/github-best-practice/best-practices.md).

## What's the context?
This change builds on #124, and can be considered a no-op.

> **Note**
> 
> 
> Some considerations:


best-practice/definitions.ts
```diff
@@ -9,6 +9,56 @@ const github: readonly IBestPractice[] = [
...
+    name: "Branch Protection",
+    owner: "[@guardian](https://github.com/orgs/guardian/teams/all)",
+    description:
+      "Enable branch protection for the default branch, ensuring changes are reviewed before being deployed.",
+    howToCheck: "Manual. View the repository on https://github.com",
+    howToExempt: "N/A",
+  },
+  {
+    name: "Access",
+    owner: "[@guardian](https://github.com/orgs/guardian/teams/all)",
```
JuliaBrigitte, line 22: What does owner @guardian mean?

best-practice/definitions.ts
```diff
@@ -9,6 +9,56 @@ const github: readonly IBestPractice[] = [
...
+      "Grant access on a team basis, rather than directly to individuals.",
+    howToCheck: "Manual. View the repository on https://github.com",
+    howToExempt: "N/A",
+  },
+  {
+    name: "Collaborators",
+    owner: "[@guardian](https://github.com/orgs/guardian/teams/all)",
+    description:
+      "Include at least one GitHub team as a collaborator with admin access.",
+    howToCheck: "Manual. View the repository on https://github.com",
```
JuliaBrigitte, line 33: Does this link to View the repository on https://github.com/ make sense?

JuliaBrigitte (approved):

mxdvl (approved):
Looks lovely

image

best-practice/definitions.ts
akash1810: > What does owner @guardian mean?

I've used it to mean "everyone", rather than a specific team.

best-practice/definitions.ts
akash1810: Hmm possibly now. Would linking to https://github.com/orgs/guardian/repositories be a decent alternative?

best-practice/definitions.ts
```diff
@@ -9,6 +9,56 @@ const github: readonly IBestPractice[] = [
...
+    howToExempt: "N/A",
```
NovemberTang, line 34: repositories that  `hackday`, `prototype`, or `learning` and no other valid status topic are exempted from needing team owners. For example in the case that a repo has both a `documentation` and `hackday` topic,`documentation` would win and a team would need to own it

best-practice/definitions.ts
```diff
@@ -9,6 +9,56 @@ const github: readonly IBestPractice[] = [
...
+  },
+  {
+    name: "Topics",
+    owner:
+      "[DevX Security](https://github.com/orgs/guardian/teams/devx-security)",
+    description:
+      "Repositories should have a topic, to help understand what is in production.",
+    howToCheck: "Manual. View the repository on https://github.com",
+    howToExempt: "N/A",
```
NovemberTang, line 51: repos owned **only** by non-p&e teams are exempt

ie
interactives -> exempt
devx & interactives -> not exempt
none -> not exempt

best-practice/definitions.ts
```diff
@@ -9,6 +9,57 @@ const github: readonly IBestPractice[] = [
...
+    description:
+      "Grant access on a team basis, rather than directly to individuals.",
+    howToCheck: "Manual. View the repository on https://github.com",
+    howToExempt: "N/A",
+  },
+  {
+    name: "Collaborators",
+    owner: "[@guardian](https://github.com/orgs/guardian/teams/all)",
+    description:
+      "Include at least one GitHub team as a collaborator with admin access.",
```
rtyley, line 32: I was surprised to see the terminology 'collaborator' here, as in the past with GitHub repos I've taken 'collaborator' to mean specifically 'outside collaborator' (accounts from outside the Guardian) - I think that's how the term was perhaps originally used in the GitHub UI? Checking some GitHub docs tho, it does seem that 'collaborator' is now used collectively for people both inside and outside of the organisation, [at least as for as the GitHub API goes](https://docs.github.com/en/rest/collaborators/collaborators?apiVersion=2022-11-28#list-repository-collaborators).

In any case, this rule might be better named 'Admin Access' or something like that - I think the key part is we want to make sure _admin_ access is granted to a team?

akash1810: Great point! We say an admin group owns the repository; 'Admin Access' or 'Ownership' are really sensible renames. Want to raise a PR?

rtyley: Done with https://github.com/guardian/recommendations/pull/134!

PR #142, opened 2023-09-13
add some exemptions

NovemberTang:
## What is being recommended?

Add some exemptions to best practice rules.

## What's the context?

We often don't need to apply a rule to a repo if it's been archived, as this indicates it's out of use.


tjsilver (approved):
👍

akash1810 (approved):

PR #145, opened 2023-09-21
new topic: interactive

NovemberTang:
## What is being recommended?

Add `interactive` as a topic for easier classification, and exclusion from relevant metrics.

Alongside this PR, we should find all repos we currently recognise as an interactive and apply this topic to them. This can be scripted

Which rules interactives are required to follow is a piece of work for another PR. We should consider whether or not we require interactive repos to be owned by a team, as their maintenance story is quite different to P+E's

## What's the context?

Currently, we have a few ways of figuring out whether a repo is an interactive or not, including checking the owner, the contributors and the repo prefix. Using a topic as a single source of truth, as we do for other kinds of repos, will simplify this process.


tjsilver (approved):
Good idea!

akash1810:
Should we also update https://github.com/guardian/recommendations/blob/main/github.md#topics?

PR #146, opened 2023-10-05
Add recommendations document for Containers

AshCorr:
## What is being recommended?

That I know of theres now 4 teams that have started using Containers in their daily work, maybe we should have some recommendations about them!

### DIFF
--- a/containers.md
+++ b/containers.md
@@ -1,0 +1,96 @@
+Containers
+======================
+
+## Installing a container runtime
+
+> [!WARNING]  
+> Do not use Docker Desktop to install Docker on MacOS. [As of Aug 2021 we need a license to use Docker Desktop at the Guardian.](https://www.docker.com/blog/updating-product-subscriptions/)
+
+MacOS doesn't support Containers natively. This unfortunately means that at the moment we need to create a Linux VM in order to run any kind of Container on our devices.
+
+Fortunately, there exists a number of tools to help us setup a VM and use it seamlessly as if we were running containers on our host machine. Docker Desktop is one of such tools, but as previously mentioned we can't use that without a license.
+
+### Podman
+
+An alternative to Docker Desktop is [Podman](https://podman.io/docs/installation). Podman is a tool built by RedHat as a complete replacement for Docker. Its completely compliant with [OCI](https://opencontainers.org/) meaning that it can run any Container that Docker can run, plus more.
+
+Podman copies most of Dockers CLI commands, so if you're used to Docker switching to Podman should be fairly straightforward, in many cases you can just create an alias from docker to podman and that'll work fine.
+
+### Minikube
+
+If you don't want to, or can't, make the switch over to Podman you can use [Minikube](https://minikube.sigs.k8s.io/docs/tutorials/docker_desktop_replacement/) instead. Like Docker Desktop, minikube also takes care of creating a VM on MacOS where containers can be ran and making it seamless to use Docker on MacOS
+
+Minikube also sets up a Development Kubernetes cluster that you can use for learning, or as an alternative to docker-compose
+
+## Building a Container Image
+
+### Use small base images
+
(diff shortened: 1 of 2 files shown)


containers.md
```diff
@@ -0,0 +1,96 @@
...
+======================
+
+## Installing a container runtime
+
+> [!WARNING]  
+> Do not use Docker Desktop to install Docker on MacOS. [As of Aug 2021 we need a license to use Docker Desktop at the Guardian.](https://www.docker.com/blog/updating-product-subscriptions/)
+
+MacOS doesn't support Containers natively. This unfortunately means that at the moment we need to create a Linux VM in order to run any kind of Container on our devices.
+
+Fortunately, there exists a number of tools to help us setup a VM and use it seamlessly as if we were running containers on our host machine. Docker Desktop is one of such tools, but as previously mentioned we can't use that without a license.
```
adamnfish, line 11: I think the recommendations repo should make a recommendation here! At the moment, is this listing options?

What are teams using already? I know we already have some Docker Desktop licences, should we double down on thay?

containers.md
```diff
@@ -0,0 +1,96 @@
...
+### Use small base images
+
+Theres thousands of base images you could choose from when building an image. Finding a good base image is one of the most important parts of your building your own image. Ideally your base image should include as little as possible, just the bare minimum to run your app, this helps reduce your final image size and also reduce the potential attack surface.
+
+ - [Alpine](https://hub.docker.com/_/alpine) is a very commonly used base image. Its extremely small image just being 5MB in size. By itself it probably doesn't have what you need, but you'll find that popular images tend to publish Alpine versions. [Node for example](https://hub.docker.com/_/node)adds the `-alpine` suffix to any images that use an Alpine base.
+ - [Distroless](https://github.com/GoogleContainerTools/distroless) Google maintains a number of images that are completely stripped down to what is absolutely necessary, they are even smaller than Alpine images. This can mean they don't even include standard system libraries that we take for granted such as libssl!
+
+Alpine is a good start for building small images, its not quite as small as Distroless but its a lot easier to work with.
```
adamnfish, line 34: Along the same lines as above, should we just say "use alpine, and choose a community maintained alpine flavour that includes your toolchain"?

containers.md
```diff
@@ -0,0 +1,96 @@
...
+
+### Avoid packaging intermediate files / code / build tools into your final Image
+
+Similar to Git, any files that get added to an image will remain in its history and can be restored even if they're deleted before the end of the build. This is to consider when optimizing your images for size as including something such as a `node_modules` folder before it has been bundled can quickly balloon your image size to be orders of magnitude larger than it needs to be.
+
+Theres 2 ways you can avoid this issue:
+  - **A**. Build your application outside of the Image build environment, and then copy your final build artifacts into the image.
```
adamnfish, line 41: I think we should try and stick with this one! In compiled languages the build and runtime toolchains are different, so this would be the normal way to do things. Let's encourage this approach everywhere, even in interpreted languages (or languages that have inherited some of the tooling and conventions of an interpreted language).

containers.md
```diff
@@ -0,0 +1,96 @@
...
+
+You may also want to consider renaming `Dockerfiles` to `Containerfiles`. 
+
+## Publishing a Container Image
+
+Instead of using the Docker container registry we recommend using [GitHubs built in Container Registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry). This registry is free to use and has no limits for any public repositories and also integrates extremely well with Github Actions.
+
+Below is an example Github workflow which handles the build and publishing of a container image to Github Container Registry.
+
+```yaml
```
adamnfish, line 56: Would it be better to link to an exaple? That way we might be more confident the example is being kept up to date?

For some of these actions we might want to use a commit hash, rather than the tag?

PR #149, opened 2023-10-30
Focus on solution over problem in Scala recommendations

mxdvl:
## What is being recommended?

Nothing new, but the Scala recommendations are refactored to focus on the solution rather that the problem.

Relegate the discussion on dependency injection to a footnote.

## What's the context?

Focus on the preferred solution over what we do not recommend. People want to solve problems, not be told off.

Original wording in #19

Pinging @guardian/scala who may want to chime in

## Images

The bullet is more consice and contains a footnote:

image

Then at the bottom of the document:

image


johnduffell (approved):
nice! sounds like the original was written by someone who had just been bitten by a hard to read codebase

PR #155, opened 2024-01-12
Copy the inclusive communication guide from our engineering culture repository

emdash-ie:
I’ve looked for [the inclusive communication guide](https://github.com/guardian/our-engineering-culture/blob/main/inclusive-communication.md) before and found it hard to find because I always look here, forgetting it’s in another repository. I think covering ~engineering culture and~ inclusive communication in our recommendations seems like a good fit, and then the info is in one place.

(Note I’ve made minor tweaks to the files in the process of copying them over, which I’ve put in a separate commit for ease of review.)

### DIFF
--- a/inclusive-communication.md
+++ b/inclusive-communication.md
@@ -1,0 +1,56 @@
+# Guidelines for inclusive communication
+
+The purpose of these guidelines are to provide some best practices and examples of inclusive communication. Just like many other best practices, you may be referred to these guidelines by your colleagues as part of code reviews, feedback sessions, document comments, and retrospectives, etc, so please do take time to digest them effectively. They are relevant to all forms of communication including verbal, written, drawing and code.
+
+Please do submit any edits you feel necessary, and these will be reviewed and merged by a member of the EM team.
+
+## Why is inclusive language important?
+
+*   Makes people feel naturally included
+*   Doesn’t trigger memories of someones’ prior trauma
+*   Brings more individuals within the perceived normality
+*   Helps overcome rather than cement peoples’ cognitive biases
+
+## Some examples
+
+#### From [Twitter Engineering on 2 July](https://twitter.com/TwitterEng/status/1278733305190342656)
+
+| **Non-inclusive** | **Alternatives** |
+|---|---|
+| Whitelist | Allowlist |
+| Blacklist | Denylist |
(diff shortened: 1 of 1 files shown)


emdash-ie:
I’ve requested review from a few people who have committed to the engineering culture repository – these are invitations rather than requests, no need to review if you don’t have time or don’t feel strongly about it!

davidfurey (commented):
I'm not convinced this is a perfect fit, in part because I don't think our culture should be considered a recommendation.

But I think the slight mismatch is worth it since merging it in will increase its visibility.

mxdvl (commented):
While I’m fully on board having our **inclusive communication** inside this repo, I’m not quite as convinced about bringing **our engineeering culture** into it. I’m not sure we’re “recommending” a culture, we foster through actions and our recommendations are means of achieving that culture.

The [CONTRIBUTING](https://github.com/guardian/our-engineering-culture/blob/main/CONTRIBUTING.md) file of our engineering culture suggests that you need 5 approvals before making a change, which is quite a different requirement than what we have for the recommendations one.

emdash-ie:
> I'm not convinced this is a perfect fit, in part because I don't think our culture should be considered a recommendation.

Ah ok, you feel it should be considered a requirement rather than a recommendation? That does make sense.

> But I think the slight mismatch is worth it since merging it in will increase its visibility.

We could possibly try another approach to increasing its visibility – a post in google chat that’s easier to find while searching, maybe, or more linking from other places. (We already have some links, so maybe the problem is just that I didn’t search for the right phrases)

emdash-ie:
> While I’m fully on board having our **inclusive communication** inside this repo, I’m not quite as convinced about bringing **our engineeering culture** into it. I’m not sure we’re “recommending” a culture, we foster through actions and our recommendations are means of achieving that culture.

I like that split – moving the inclusive communication guide into this repo solves my problem of having trouble finding it, and I agree that I feel more comfortable calling that a recommendation than the culture.

> The [CONTRIBUTING](https://github.com/guardian/our-engineering-culture/blob/main/CONTRIBUTING.md) file of our engineering culture suggests that you need 5 approvals before making a change, which is quite a different requirement than what we have for the recommendations one.

Ah yes, I meant to copy over this file as well in some way (or at least compare the equivalent for this repo) but forgot.

mxdvl (approved):

PR #156, opened 2024-01-25
Adds a linter for inclusive language suggestions

NovemberTang:
## What's the change?

[Alex](https://alexjs.com/) is a configurable linter for catching exclusionary language in documentation. I've applied it to this repo as a potential way of automating our [inclusive communication guide](https://github.com/guardian/our-engineering-culture/blob/main/inclusive-communication.md).

## What's the impact

Typically, CI would fail if alex flags any warning or errors, but for now, I'm catching all non-zero exit codes so we don't prevent new contributions until the backlog is cleared. Please view the latest action run to see that kind of language it is looking for.

Alex catches uses of words that people may consider offensive or generally not inclusive, such as words with gendered or racial connotations, or language that assumes a particular level of experience. For example, in the sentence `Just spin up a container.`, the word `just` is doing a lot of heavy lifting, as it's not a trivial task for everyone. Alex catches these and suggests alternatives, or removing the word outright (fwiw this is definitely something I catch myself doing, and it's been useful to have a reminder that my skillset is not universal).

Example output
```
elasticsearch.md
    15:100-15:106  warning  `master` may be insensitive, use `primary`, `lead`, `hub`, `reference` instead        master   retext-equality

github.md
    24:267-24:271  warning  `just` may be insensitive, try not to use it                                          just     retext-equality

```

If it detects any warnings or errors, it will return a non-zero exit code. When running the GH Action, I've caught this error so that it doesn't prevent ongoing contribution to the repo while the backlog is being worked through.

This PR could be a follow on to #155 , raised by @emdash-ie 

## Configuration

Alex can be configured to ignore certain rules. I've set it to exclude references to stewards/stewardesses, so we don't catch Scala Steward references. I've also switched the profanitySureness setting from 0 to 1 (the maximum is 2), to make the warnings less noisy.


.alexrc.js
```diff
@@ -0,0 +1,4 @@
+exports.profanitySureness = 1;
+exports.allow = [
+    "steward-stewardess"
```
NovemberTang, line 3: We have a lot of references to Scala Steward, so I've excluded this rule.

akash1810:
A brilliant addition!

Reporting of `master` in the context of Elasticsearch, whilst correct, will be tricky to resolve as it's [terminology used by Elastic themselves](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-node.html).

akash1810 (approved):

mxdvl (commented):
This looks amazing!

We’ve got a [very minimal denylist in DCR](https://github.com/guardian/dotcom-rendering/pull/5879) which could be greatly improved 👀

emdash-ie (approved):

NovemberTang:
> Also, checking a few of the warnings in the build on this PR they look like false positives to me:

Good spot, I've updated the ignore list to exclude `special`. I reckon the other `just` instances are simple enough that we can get by without them, but I do take your point that it's not very precise

PR #158, opened 2024-02-05
Link to CNAME migration docs

bryophyta:
## What is being recommended?

Adds a link to the CDK docs on migrating an existing NS1 record to CDK.

## What's the context?

This is salient information for people who want to understand and implement our existing DNS recommendations, but currently is somewhat difficult to discover because it's included as a step in the migration guide. (I don't think it's separately documented elsewhere, but could be mistaken.)

### DIFF
--- a/domain-names.md
+++ b/domain-names.md
@@ -2,7 +2,9 @@
 
 ## Which DNS provider should I use?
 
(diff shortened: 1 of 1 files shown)


tjsilver (approved):
Looks good, thank you!

PR #160, opened 2024-03-14
Add recommendation for specifying Github Actions versions

chrislomaxjones:
## What is being recommended?

Add a recommendation for specifying versions of Github Actions in our workflows.

As there wasn't a recommendation for GHA in general, I've added an intentionally brief intro. This could be fleshed out in the future, as required.

## What's the context?

This approach was reported by the Coverdrop team (who use it in their project).

Feedback welcome!


github-actions.md
```diff
@@ -0,0 +1,24 @@
...
+- Reference via a Git tag e.g `uses: actions/checkout@v4`. These correspond to tagged releases on Github - see [actions/checkout/releases](https://github.com/actions/checkout/releases) for examples.
+- Reference via a Git Commit SHA e.g. `uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1`.
+
+**We recommend specifying the version of Github Actions workflows as a commit SHA**.
+
+A security issue arises when using tags to specify versions: code executed by the underlying action can be changed without creating a new release. A malicious actor could exploit this in order to run code within our own workflows. See this [blog post](https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash) for a complete example.
+
+Since commit SHAs are immutable, the code of the underlying workflow cannot be changed for any given commit. This mitigates a security issue that arises from using tags (assuming you’re comfortable with the code present for the given SHA).
+
+It’s important to also audit the action itself for the given commit, so you’re reasonably satisfied the action is behaving as expected.
```
itsibitzi, line 18: It might make sense to clarify what "audit" means in this instance. Hopefully most folks would be able to infer what it means but I don't think it hurts to be super explicit? I think we mean that a developer should go and manually read the action to make sure it doesn't do things like exfiltrate secrets?

chrislomaxjones: Yup, good shout.

github-actions.md
```diff
@@ -0,0 +1,24 @@
...
+Github actions should be used for Continuous Integration. Actions are typically used to enforce checks before merging branches into main (such as formatting, linting and testing), as well as to publish libraries or [deploy](continuous-deployment.md) build artifacts (via Riff-Raff).
+
+## Specifying versions for actions you use
+
+Github Actions workflows can invoke other actions via `uses` steps. When specifying the version to use, one can typically:
+
+- Reference via a Git tag e.g `uses: actions/checkout@v4`. These correspond to tagged releases on Github - see [actions/checkout/releases](https://github.com/actions/checkout/releases) for examples.
+- Reference via a Git Commit SHA e.g. `uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1`.
+
+**We recommend specifying the version of Github Actions workflows as a commit SHA**.
```
itsibitzi, line 12: Perhaps we should include some information on how to find the SHA digest for a particular release?

I personally go to the releases page and click the SHA digest in the panel on the left (below) then copy the full commit SHA out of the URL. This might not be the best way though. 

image

github-actions.md
```diff
@@ -0,0 +1,24 @@
...
+
+A security issue arises when using tags to specify versions: code executed by the underlying action can be changed without creating a new release. A malicious actor could exploit this in order to run code within our own workflows. See this [blog post](https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash) for a complete example.
+
+Since commit SHAs are immutable, the code of the underlying workflow cannot be changed for any given commit. This mitigates a security issue that arises from using tags (assuming you’re comfortable with the code present for the given SHA).
+
+It’s important to check the source code of the action for the given commit, so you’re reasonably satisfied the action is behaving as expected (and not exfiltrating secrets, for example).
+
+As well as specifying the commit, it’s worth combining this with a comment specifying a more readable semver version. Dependabot also knows how to handle updates for workflows versioned with SHAs, with a comment that is kept updated with the version tag that the commit points to: see [nodejs/node/pull/51334](https://github.com/nodejs/node/pull/51334) for an example.
```
tjsilver, line 20: Maybe ... "... a comment with the semver version in the format ` # v.0.0.1`",  or similar? Just to make it clear that Dependabot will update it in this particular format.

chrislomaxjones: Nice idea

github-actions.md
chrislomaxjones: Good idea, I'll nab your screenshot while I'm at it

github-actions.md
```diff
@@ -0,0 +1,30 @@
...
+Github Actions workflows can invoke other actions via `uses` steps. When specifying the version to use, one can typically:
+
+- Reference via a Git tag e.g `uses: actions/checkout@v4`. These correspond to tagged releases on Github - see [actions/checkout/releases](https://github.com/actions/checkout/releases) for examples.
+- Reference via a Git Commit SHA e.g. `uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1`.
+
+**We recommend specifying the version of Github Actions workflows as a commit SHA**.
+
+A security issue arises when using tags to specify versions: code executed by the underlying action can be changed without creating a new release. A malicious actor could exploit this in order to run code within our own workflows. See this [blog post](https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash) for a complete example.
+
+Since commit SHAs are immutable, the code of the underlying workflow cannot be changed for any given commit. This mitigates a security issue that arises from using tags (assuming you’re comfortable with the code present for the given SHA).
```
NovemberTang, line 16: minor: I would swap lines 14 and 16 around, as I think it's easier to understand this way.

github-actions.md
```diff
@@ -0,0 +1,30 @@
...
+
+### Finding the SHA for a given release
+
+You can find the SHA for a particular release by navigating to the Releases page of a GitHub repository, and clicking the short SHA digest in the panel to the left of the release. The full SHA can then be copied out of the URL.
+
+![image](https://private-user-images.githubusercontent.com/5560113/312924049-ed6f7db5-7b89-49ac-acd4-7287ef274335.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTA0MzYwMzAsIm5iZiI6MTcxMDQzNTczMCwicGF0aCI6Ii81NTYwMTEzLzMxMjkyNDA0OS1lZDZmN2RiNS03Yjg5LTQ5YWMtYWNkNC03Mjg3ZWYyNzQzMzUucG5nP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MDMxNCUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDAzMTRUMTcwMjEwWiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9NTk5YTRmMzg1NDhmMGU0ZGYzMDk3MGZmOWRmODY4NjJhNzE0OWY3NDIzNGFhZDljZjNlZTRmZTc3N2YwNTk3MCZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmYWN0b3JfaWQ9MCZrZXlfaWQ9MCZyZXBvX2lkPTAifQ.xTHUwbsDJZ5XqTY39f9BlQxefou4myvNqxbvsaW8isc)
```
NovemberTang, line 26: The image looks like it might be broken
Screenshot showing a broken image link

github-actions.md
```diff
@@ -0,0 +1,30 @@
...
+
+Since commit SHAs are immutable, the code of the underlying workflow cannot be changed for any given commit. This mitigates a security issue that arises from using tags (assuming you’re comfortable with the code present for the given SHA).
+
+It’s important to check the source code of the action for the given commit, so you’re reasonably satisfied the action is behaving as expected (and not exfiltrating secrets, for example).
+
+As well as specifying the commit, it’s worth combining this with a comment specifying a more readable semver version, in the format `# v0.0.1`. Dependabot also knows how to handle updates for workflows versioned with SHAs, with a comment that is kept updated with the version tag that the commit points to: see [nodejs/node/pull/51334](https://github.com/nodejs/node/pull/51334) for an example.
+
+### Finding the SHA for a given release
+
+You can find the SHA for a particular release by navigating to the Releases page of a GitHub repository, and clicking the short SHA digest in the panel to the left of the release. The full SHA can then be copied out of the URL.
```
akash1810, line 24: This can also be scripted, for example, to get the releases for https://github.com/actions/checkout:

```sh
git ls-remote --tags https://github.com/actions/checkout | sort -Vr -k2
```

github-actions.md
```diff
@@ -0,0 +1,30 @@
...
+## Specifying versions for actions you use
+
+Github Actions workflows can invoke other actions via `uses` steps. When specifying the version to use, one can typically:
+
+- Reference via a Git tag e.g `uses: actions/checkout@v4`. These correspond to tagged releases on Github - see [actions/checkout/releases](https://github.com/actions/checkout/releases) for examples.
+- Reference via a Git Commit SHA e.g. `uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1`.
+
+**We recommend specifying the version of Github Actions workflows as a commit SHA**.
+
+A security issue arises when using tags to specify versions: code executed by the underlying action can be changed without creating a new release. A malicious actor could exploit this in order to run code within our own workflows. See this [blog post](https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash) for a complete example.
```
akash1810, line 14: Is it worth also noting this recommendation differs from [GitHub's](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions)?

IMO GitHub's recommendation strikes a better balance between DX and security, though admittedly is more difficult to enforce (once we get there!).

github-actions.md
```diff
@@ -0,0 +1,30 @@
...
+
+Since commit SHAs are immutable, the code of the underlying workflow cannot be changed for any given commit. This mitigates a security issue that arises from using tags (assuming you’re comfortable with the code present for the given SHA).
+
+It’s important to check the source code of the action for the given commit, so you’re reasonably satisfied the action is behaving as expected (and not exfiltrating secrets, for example).
```
akash1810, line 18: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-openssf-scorecards-to-secure-workflows might be a way to automate these checks?

github-actions.md
itsibitzi: GitHub seem to also recommend pinning on the commit digest?

github-actions.md
itsibitzi: Would rely on the 3rd party action having a score card on their repository, and them not spoofing it, I suppose.

I think the scorecards are a really good thing, but I'm not sure they're a substitute for auditing if we're to take security seriously. We could possibly have a repository of manually audited actions and their hashes? `cargo vet` does something like this.

github-actions.md
akash1810: I was referring to the "Pin actions to a tag only if you trust the creator" section. I'd suggest we're able to trust actions authored by ourselves, GitHub, AWS, etc. Actions from third-party individuals, such as [EnricoMi/publish-unit-test-result-action](https://github.com/guardian/content-api/blob/4cb34b2ee185203ac5ca059c0d08615ed2c8a8c3/.github/workflows/porter.yml#L71) would be pinned.

itsibitzi: Personally I think the principle of zero-trust applies here. Sure, I trust no one at the Guardian would maliciously attempt to steal my repo secrets, but if there's a compromised GitHub action repo then I'd prefer if that didn't cascade to my secrets being stolen.

If dependabot is able to flag updates with commit digests then I think we should be able to apply updates with a small amount of toil (I generally don't want random, unrequested updates landing in my codebases regardless of the security implications).

akash1810: > If dependabot is able to flag updates with commit digests then I think we should be able to apply updates with a small amount of toil

Agree in principle. However in practice, I think the story is a little different as:
- Usage of Dependabot to bump GHA isn't universally applied
- Timely review of Dependabot PRs is also not universal; there are a couple thousand [PRs open today](https://github.com/pulls?q=is%3Aopen+is%3Apr+org%3Aguardian+author%3Aapp%2Fdependabot)

The volume of open PRs is probably a signal of some missing tooling or processes TBH, but that's a different problem to solve...

github-actions.md
chrislomaxjones: Thanks, have swapped around and tweaked the wording so it reads slightly better.

github-actions.md
chrislomaxjones: Fixed now.

github-actions.md
chrislomaxjones: Very cool, will add to the recommendation.

github-actions.md
chrislomaxjones: These scorecards and manual auditing sound good, but I'm hesitant to add to this recommendation as it stands, just because I haven't encountered them before. Maybe we can do some investigation in the future and update the recommendation down the line?

github-actions.md
chrislomaxjones: Thanks for the feedback both. I'll leave the recommendation as it is, but mention that we're taking a slightly different stance to the one that is recommended by Github.

NovemberTang (approved):
Thanks for all the work to get this over the line @chrislomaxjones

PR #161, opened 2024-03-14
Update wording to reflect teams' shared responsibility

adamnfish:
## What is being recommended?

In a no-blame culture, all the work we produce is a shared responsibility. This means if a change causes an incident, it isn't the fault of the engineer that made the change. This cuts both ways though, if delivering software is a shared responsibility then this means reviewers aren't completely absolved.

## What's the context?

I believe this change better reflects the reality of how teams work here today. With the rise of AI-assistants, we will need to instill a renewed focus on code review and I believe that even more than now, a shared responsibility model for teams will be essential.

This change serves as a trigger for checking that my sense of how this works is shared, ahead of rolling out guidelines for the use of AI assisted programming.


NovemberTang (approved):

bryophyta (approved):
this sounds good to me, and from my experience I think is the approach that I've tended to see teams adopting in the department

PR #162, opened 2024-03-14
Upgrade version of Java from 11 to 21

mchv:
## What is being recommended?

Update the version of Java to the latest LTS. We need to update the associated doc as well, but at least this clarify the expectation.


johnduffell (approved):
The doc is mostly updated in structure, of course as people learn more things they can add more examples and learnings.

PR #165, opened 2024-04-15
docs: update ruleset instructions

tjsilver:
## What is being recommended?

Updates the ruleset recommendations for CI status checks with instructions for when a workflow has more than one job in it. In that case, the job names should be used as the workflow name will fail.

## What's the context?
Rulesets can be used to re-apply branch protections when branch protection must be turned off on a repo. Branch protections applied via Ruleset can exclude certain apps, such as the Gu Scala Library Release app (which is not possible with 'normal' branch protection).


github-rulesets.md
```diff
@@ -6,12 +6,15 @@ An example of this is the [GHA Scala Library Release Workflow](https://github.co
...
 
 To require CI checks to pass before merge:
 * Create a second ruleset in your repo with the 'Branch protection' property 'Require status checks to pass' -> 'Require branches to be up to date before merging'.
-* Type your repo's workflow name(s) (not to be confused with the filename) into the 'Status checks that are required' box. For example, if your repo has a CI workflow with the name 'CI', then the setting would look like this:
+* Type your repo's workflow name(s) (not to be confused with the filename) into the 'Status checks that are required' box. (If your CI workflow has more than one job listed, list the job names instead.) 
+  * For example, if your repo has a CI workflow with the name 'CI', then the setting would look like this:
 ![status_checks.png](images/status_checks.png)
+  * Or if your workflow has several jobs called 'build', 'trigger-workflow' and 'check-status', it would look like this:
```
adamnfish, line 16: Are we sure this is the case? Is it that we should always be writing the job name, and it's just that by chance people tended to pick the same names for the job and the workflow?

If this is true and there is a different requirement depending how many steps there are then no worries, although we might be able to make this a bit clearer by having the top level bullet point say something a bit more general "refer to the build workflow, depending how many steps it has", and then provide the two different options in the nested bulletpoints.

My fear is that the top level bullet point at the reads like it used to say "type the workflow name" and then got changed to include the point about multiple steps! It slightly leads someone towards writing the workflow name and then later mentions the condition about whether this is the right approach!

adamnfish (approved):
Great!

PR #170, opened 2024-06-25
Reference the whole WCAG standard

bryophyta:
## What is being recommended?

Recommending the whole WCAG AA standard as our benchmark. I believe that we do currently see ourselves as accountable to WCAG AA in the department, and that we're right to do so.

If I understand correctly then the original reason for mentioning 'a subset' was to prevent the recommendation from feeling too overwhelming for readers who are new to the standard (which is quite detailed, and not always the most user-friendly for newcomers). In light of this, I've explicitly reframed the list of core principles as a starting point, which should hopefully point people in the direction of the most impactful parts of the spec, if they're new to the topic.

## What's the context?

> **Note**
> 
> 
> Some considerations:

### DIFF
--- a/accessibility.md
+++ b/accessibility.md
@@ -2,22 +2,19 @@
 
 ## Compliance
 
-We use a subset of the WCAG 2.1 AA standards as our benchmark for accessible products.
+We use compliance with WCAG 2.1 AA standards as our benchmark for accessible products.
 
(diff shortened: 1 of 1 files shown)


SiAdcock (approved):
👍 Great change, thanks @bryophyta

PR #174, opened 2024-07-12
Scheduling behaviour for multi-instance apps

bryophyta:
## What is being recommended?

Leveraging an app's load balancer to schedule tasks with side-effects on multi-instance apps.

## What's the context?

This is a problem that my team and I have come across a few times, and describes a solution that we've found to work well in production a few times.

I presented on this approach at Server Side Meetup a couple of months ago and it was generally well-received, and I've incorporated some extra considerations that were raised in discussion into this recommendation.

### DIFF
--- a/scheduled-jobs.md
+++ b/scheduled-jobs.md
@@ -21,7 +21,7 @@ See also https://opensource.com/article/20/7/systemd-timers.
 
 ### Multi-instance apps (e.g. EC2 in an auto-scaling group)
 
-`crontab` and similar solutions won't always be suitable for multi-instance apps. For example, if you have more than one instance but only need to run the scheduled task once (common if the task has side effects, like sending email).
+`crontab` and similar solutions won't always be suitable for multi-instance apps. For example, if you have more than one instance but only need to run the scheduled task once (common if the task has side effects, like sending email).[^1]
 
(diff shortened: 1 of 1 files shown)


mchv:
This looks like a great addition to me on a very practical problem, and this has already has been presented to a wider audience. It would be great to have someone from another team who have use this pattern to quickly review before we merge?

mchv (approved):

bryophyta:
> It would be great to have someone from another team who have use this pattern to quickly review before we merge?

Thanks for this suggestion. I'm not sure whether other teams are currently using this pattern.. 

The `EventBridge Rule + ApiDestination` pattern is being used in the [`fastly-content-fanout`](https://github.com/guardian/fastly-content-fanout/blob/26db86105f3022d7a220e3c0f785ff62265a82d6/cdk/lib/eventbridge-to-fanout.ts#L40) project. From discussion with @twrichards (currently on leave) I believe that they saw a very similar set of pros and cons when using this pattern, although the use-case was slightly different (scheduling calls to a 3rd party service, hence not so concerned with the multi-instance app issue as such).

But if we'd like to see multiple teams using a solution before adding it to this repo then I can pause this PR for now?

PR #177, opened 2024-07-31
Note that +compile should be used in CI if crossScalaVersions is set

emdash-ie:
## What's the context?

> **Note**
> 
> 
> Some considerations:


mchv (approved):

emdash-ie:
@mchv Thanks for the approval! I had forgotten about this PR – will merge now.

I think the need to make the change from just to only here to satisfy the linter is silly (i.e. I think the linter should be changed), but I don't feel strongly enough about it to look into it now.

PR #183, opened 2025-02-12
clarify scala version recommendation

johnduffell:
Now that scala has LTS and next, we should follow the official recommendation for our code.
https://www.scala-lang.org/development/

I also removed a couple of lines that are probably just noise nowadays.

### DIFF
--- a/scala.md
+++ b/scala.md
@@ -1,9 +1,7 @@
 # Scala
 
- * Always specify an explicit sbt.version
- * Ensure you run in production with GC logging enabled
  * Prefer immutability
- * Prefer the latest version of Scala. At the time of writing `3.1.2` is the latest version, so `2.13.8` may be acceptable. if you are still using `2.12.x` you should plan to migrate.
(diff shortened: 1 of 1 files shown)


scala.md
```diff
@@ -1,9 +1,7 @@
 # Scala
 
- * Always specify an explicit sbt.version
```
johnduffell, line 3: intellij generates one by default

scala.md
```diff
@@ -1,9 +1,7 @@
...
- * Ensure you run in production with GC logging enabled
```
johnduffell, line 4: I assume we are not putting enough new services into production to warrant this being so prominent.  Now we should probably add more recommendations about maintainability e.g. ScalaStyle and library use.  Perhaps one for the scala guild to consider.

kelvin-chappell: GC logging without some idea of what you're going to do with it doesn't seem that useful.

kelvin-chappell (approved):
👍

PR #185, opened 2025-06-18
Add suggestion to self-review your PRs

johnduffell:
This PR adds a recommendation to review and comment on your own PRs.

This is practised by a few people in supporter revenue and seems to work well for the person raising the PR and the reviewers, so it should be considered by everyone.

### DIFF
--- a/pull-requests.md
+++ b/pull-requests.md
@@ -50,6 +50,8 @@ If these things are not true then you should address the blockers before raising
 ## Descriptive
 A PR's description should include sufficient information on what the change is and why it is necessary.
 
+You should self-review your PR and leave comments to pre-empt any questions that are likely to come up at the right moment, or draw attention to areas that need more attention.
+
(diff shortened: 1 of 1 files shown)


shtukas:
+1, and thanks! I am surprised it wasn't already there :)

pull-requests.md
```diff
@@ -50,6 +50,8 @@ If these things are not true then you should address the blockers before raising
 ## Descriptive
 A PR's description should include sufficient information on what the change is and why it is necessary.
 
+You should self-review your PR and leave comments to pre-empt any questions that are likely to come up at the right moment, or draw attention to areas that need more attention.
+
```
emdash-ie, line 54: I like having this suggestion, but I wonder would it be better placed down a couple of paragraphs? Here it interrupts the flow of what was there, meaning that the ”future peace of mind” bit now refers to the self-review, where it used to refer to having a PR at all. I would be tempted to place it as the last paragraph under this heading.

emdash-ie: (sorry I didn’t get to this before merge!)

bryophyta:
i realise it's already merged but i find this practice really useful too -- great addition!

PR #199, opened 2026-04-01
explain why dependency pinning is important

NovemberTang:
## What is being recommended?

In light of the recent increase in supply chain attacks, we should expand our dependency recommendations, explaining why not pinning your dependencies to the minor version presents a security risk

## What's the context?

https://snyk.io/blog/axios-npm-package-compromised-supply-chain-attack-delivers-cross-platform/

### DIFF
--- a/dependencies.md
+++ b/dependencies.md
@@ -39,8 +39,8 @@ This provides more [consistency and safety between the `package.json` file and t
 
 Why not pinning dependencies is dangerous:
 - You are more vulnerable to supply chain attacks, as you are pulling in the
-latestversion of a dependency, which may have been compromised. Explicitly
-raising a PR (bonus points you do this using an automated system with a
+latest version of a dependency, which may have been compromised. Explicitly
+raising a PR (bonus points if you do this using an automated system with a
 [cooldown period](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-))
 often results in a delay of a few hours to several days, which is usually
 enough time for the maintainers, or package manager to detect security
--- a/dependencies.md
+++ b/dependencies.md
@@ -35,7 +35,24 @@ When using a package manager, use a lock file to prevent prevent mismatches in t
 dependencies between environments.
 
 Ensure that all dependencies in deployed applications are pinned to a `patch` version.
-This provides more [consistency and safety between the `package.json` file and the lock file](https://docs.renovatebot.com/dependency-pinning/#what-a-lock-file-doesnt-do-for-you).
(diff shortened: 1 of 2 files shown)


dependencies.md
```diff
@@ -35,7 +35,24 @@ When using a package manager, use a lock file to prevent prevent mismatches in t
 dependencies between environments.
 
 Ensure that all dependencies in deployed applications are pinned to a `patch` version.
-This provides more [consistency and safety between the `package.json` file and the lock file](https://docs.renovatebot.com/dependency-pinning/#what-a-lock-file-doesnt-do-for-you).
+This provides more [consistency and safety between the `package.json` file and the lockfile](https://docs.renovatebot.com/dependency-pinning/#what-a-lock-file-doesnt-do-for-you).
+
+Why not pinning dependencies is dangerous:
+- You are more vulnerable to supply chain attacks, as you are pulling in the
+latestversion of a dependency, which may have been compromised. Explicitly
```
andrewHEguardian:
```suggestion
latest version of a dependency, which may have been compromised. Explicitly
```

dependencies.md
```diff
@@ -35,7 +35,24 @@ When using a package manager, use a lock file to prevent prevent mismatches in t
...
+raising a PR (bonus points you do this using an automated system with a
```
andrewHEguardian:
```suggestion
raising a PR (bonus points if you do this using an automated system with a
```

andrewHEguardian (approved):
This is great justification and explanation! Found two tiny typos

SiAdcock:
Is there a case to be made here for a distinction between dependency version pinning in applications vs libraries intended to be published to NPM?

I absolutely agree applications should pin exact versions for reasons mentioned, and I can't think of any downsides.

There is at least one good reason why you might want to specify dependency version ranges in libraries, rather than pinning exact versions for all dependencies. Allowing the package manager to resolve dependency ranges means we can install the minimum number of dependencies and avoid bloating `node_modules` with lots of almost-duplicated dependencies. Is it worth having that discussion here?

See chat: https://chat.google.com/room/AAAAWwBdSMs/RKzDgOP8p3E/RKzDgOP8p3E?cls=10

NovemberTang:
> Is there a case to be made here for a distinction between dependency version pinning in applications vs libraries intended to be published to NPM?

@SiAdcock  It looks like this is already happening in [npm-packages.md](https://github.com/guardian/recommendations/blob/main/npm-packages.md). It's maybe worth revisiting the advice in that file at some point to make sure it's still best practice, but for now, I'll make it clearer that this applies mainly to applications.

PR #200, opened 2026-04-27
Add permission recommendations for GHA workflows

kelvin-chappell:
## What is being recommended?
Permissions should be set explicitly in GHA workflows so that the scope permitted to a job is clear.

## What's the context?
A query about best practices for using third-party actions.


tjsilver (approved):
Very clear!

PR #204, opened 2026-06-04
Recommend GitHub Rulesets over Branch Protection Rules

rtyley:
## What is being recommended?

Use GitHub Rulesets over Branch Protection Rules to provide branch protection on repos. Any new repo should use rulesets for preference.

## What's the context?

@davidfurey used a bot to search for inconsistencies in our recommendations repo, which prompted him to ask why:

> ...we only use Rulesets for branch protection on repos using GHA Scala Library Release Workflow rather than using it for all repos?

The truth is that it is [_required_ for repos using GHA Scala Library Release Workflow](https://github.com/guardian/gha-scala-library-release-workflow/blob/main/docs/configuration.md#branch-protection), but probably also a good idea for other repos too:

* more consistent
* allows repos to take advantage of our org-wide rulesets
* Rulesets are more configurable and have capabilities like GitHub App restrictions bypass that Branch Protection Rules do not have

### DIFF
--- a/github-rulesets.md
+++ b/github-rulesets.md
@@ -1,23 +1,34 @@
 # GitHub Rulesets
 
-Rulesets can be used to apply branch protection and CI status check rules to a repo, in the case where branch protection needs to be bypassed in certain cases. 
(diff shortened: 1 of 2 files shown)


davidfurey (approved):
Thanks Roberto!

PR #206, opened 2026-07-21
make cfnable domains clearer

johnduffell:
The docs were not clear that theguardian.com isn't cfnable.

This PR makes it clear.


domain-names.md
```diff
@@ -4,7 +4,7 @@
 
 NS1 is our preferred supplier for DNS hosting. We pay for their dedicated DNS service, which is independent from their shared platform.  This means that even if their shared platform experiences a DDOS attack, our DNS will still be available.
```
johnduffell, line 5: you can display the "rich diff" above, 

Image

or look at the preview here:
https://github.com/guardian/recommendations/blob/jd/clearer-docs/domain-names.md

davidfurey (approved):
Thanks for the tip on rich diffs, looks good to me.

PR #207, opened 2026-08-06
docs: Provide worked examples for the AWS tag schema

akash1810:
## What is being recommended?
Updates the recommendation for AWS tags, providing a worked example of our tagging schema.

> [!NOTE]
> It might be easier to review the [rendered markdown](https://github.com/guardian/recommendations/blob/99be17fe0ac4be73f9395989d377183565b3c8eb/aws-tags.md).

## What's the context?
The existing definition of `stack` is quite ambiguous. Hopefully, the worked examples is clearer.


aws-tags.md
```diff
@@ -0,0 +1,50 @@
+# AWS Tags
+
+Most resources in AWS can be [tagged](https://docs.aws.amazon.com/whitepapers/latest/tagging-best-practices/what-are-tags.html).
+This document outlines tags used at the Guardian. In short, tag resources with:
+- `App`
+- `Stack`
+- `Stage`
+- `gu:repo`
```
adamnfish-gu, line 8: I don't think people should be setting this manually? Is it better to omit this, and have it happen when people use CDK? Via CDK these get automatically propagated to the stack's resources, so adding this asa. recommendation is probably going to encourage people to erroneously start adding this tag everywhere.

aws-tags.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Core tags
+The following tags should be applied to all taggable resources.
+
+### `App`
+This tag identifies an individual application.
+
+For example, `user-api` could be a web service for managing user's preferences, or `user-cleanup` could be a lambda which runs periodically to remove inactive users.
+
+### `Stack`
```
adamnfish-gu, line 20: I think it'd be nice to call out the implications of this tag as well, to help people understand its use. They are at least:

- Riff Raff uses it to work out which AWS account to deploy to, via a manually curated list of mappings
- Anghammarad can use the Stack to route messages to a team for the entire group of application, which can be easier than adding an Aghammarad mapping for every app individually.

aws-tags.md
```diff
@@ -0,0 +1,50 @@
...
+> "Stack" is an overloaded term. For example, in some contexts it can mean a CloudFormation stack.
+> Indeed, a CloudFormation stack should have a `Stack` tag!
+
+### `Stage`
+This tag identifies the environment. Typical values are:
+- `PROD` for production
+- `CODE` for pre-production
+- `INFRA` for account-wide infrastructure or singleton resources e.g. [elasticsearch-node-rotation](https://github.com/guardian/elasticsearch-node-rotation)
+
+### `gu:repo`
```
adamnfish-gu: As above, I think this is an separate concept. We have a bunch of automated metadata, including this one. Perhaps here we should focus on the tags that teams need to explicitly think about?

aws-tags.md
```diff
@@ -0,0 +1,50 @@
...
+> [!IMPORTANT]
+> The combination of `App`, `Stack`, `Stage` should **uniquely identify** a service.
+> For example, there will be one lambda tagged `App=user-cleanup`, `Stack=user-management`, `Stage=PROD` in the entire estate.
+
+## Other common tags
+There are some additional tags to consider based on the circumstance.
+
+### `Owner`
+When provisioning a resource in another team's account, the `Owner` tags helps that team know who to contact if needed. 
+It should be an email address.
```
adamnfish-gu: Is this true?! Seems like a strange choice! Mailing lists change when teams change, and having email addresses lying around in infrastructure metadata feels likely to cause issues or confusion down the line.

AWS.md
```diff
@@ -3,14 +3,7 @@
...
    * Prefer to [import resources](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-import.html) into CFN over manually configuring them
- * Tag resources with:
-   * `Stack` - the broad umbrella the service sits under e.g. `media-service`. Useful to denote ownership within a shared AWS account.
-   * `Stage` - the environment, typical values are:
-     * `PROD` for production
-     * `CODE` for pre-production
-     * `INFRA` for infrastructure or singleton resources e.g. [elasticsearch-node-rotation](https://github.com/guardian/elasticsearch-node-rotation)
-   * `App` - the individual service e.g. `image-loader`
-   * `gu:repo` - the GitHub repository where the resource's definition can be found
+ * Resources should be tagged. See [aws-tags.md](./aws-tags.md) for more information.
```
adamnfish-gu, line 6: Great 👍

aws-tags.md
akash1810: Oops yeah you're right; we don't use email, but (s)team names - https://github.com/search?q=org%3Aguardian+%22add%28%27Owner%27%22+language%3Atypescript+path%3A**%2Fcdk%2F**&type=code.

AWS.md
```diff
@@ -3,14 +3,7 @@
...
  * If a resource needs to be shared across multiple environments, prefer to define it in it's own CFN template as the same resource cannot be defined in multiple templates
```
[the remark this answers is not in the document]
akash1810: Out of scope for this PR.

aws-tags.md
akash1810: [Riff-Raff will also add `gu:repo`](https://github.com/guardian/riff-raff/blob/5c7b9dc8612d21f13e93f5e03c3fccda5925435e/riff-raff/app/AppComponents.scala#L114); you're right though, distinguishing between user set tags and machine set tags would be useful.

aws-tags.md
```diff
@@ -0,0 +1,50 @@
+# AWS Tags
+
+Most resources in AWS can be [tagged](https://docs.aws.amazon.com/whitepapers/latest/tagging-best-practices/what-are-tags.html).
+This document outlines tags used at the Guardian. In short, tag resources with:
+- `App`
+- `Stack`
+- `Stage`
+- `gu:repo`
+
+You can apply your own tags in addition too.
```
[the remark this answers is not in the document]
akash1810: Updated.

aws-tags.md
akash1810: I've added a note to say `gu:repo` is automatically set by tooling.

aws-tags.md
akash1810: Added.

aws-tags.md
akash1810: Fixed.

shtukas (approved):

johnduffell (approved):
this is great, I think we don't really use stack properly in supporter revenue, partly as we just have loads and loads of glue lambdas and only a small number of larger systems.

akash1810:
This repository should be understood by a new starter within their first week of joining (a slightly made up metric!). I'm unsure if this change meets that requirement, however as its an improvement to the current documentation, I'll merge and look to refactor over time.

